PR #76, opened 2016-01-19
Adds "Translate All" example widget

DanweDE:
This widget translates text of all the entry's fields from the default
locale to all other locales enabled in the space. It uses the Yandex
Translation API.

[TP # 7703](https://contentful.tpondemand.com/entity/7703)

### DIFF
--- a/examples/translate-all/README.md
+++ b/examples/translate-all/README.md
@@ -1,0 +1,39 @@
+# “Translate All Content” Sidebar Widget
+
+This widget translates text of all the entry's fields from the default locale to all
+other locales enabled in the space, using the
+[Yandex translation API](https://translate.yandex.com/developers).
+
+## Bootstrap example for local development
+
+In the directory containing this README, run
+```bash
+npm install
+```
+
+First set the access token on your environment, then create the widget on Contentful:
+```bash
+export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=<contentfulManagementApiToken>
+contentful-widget create --space-id <yourSpaceId>
+```
+
+Serve on `http://localhost:3000`:
+```bash
+gulp watch
+```
+This does automatically serve again on any change to the source files, so you only
+need to manually refresh your browser to see the changes.
+
+**Note:** Since the Contentful App is served through HTTPS while the widget code is
+loaded from your local machine through HTTP, you need to enable insecure content.
+See here how to do it in [Firefox][ff-mixed] and [Chrome][chrome-mixed].
+
+[ff-mixed]: https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox
+[chrome-mixed]: https://support.google.com/chrome/answer/1342714
+
+## Upload widget
+If you want to inline all dependencies and upload the widget entirely to Contentful:
+```bash
+gulp bundle
+contentful-widget update --srcdoc ./dist/index.min.html --space-id <yourSpaceId> --force
+```
--- a/examples/translate-all/README.md
+++ b/examples/translate-all/README.md
@@ -6,8 +6,15 @@ other locales enabled in the space, using the
 
 ## Bootstrap example for local development
 
-In the directory containing this README, run
+In the *widget-sdk* directory root, build the widget API with
 ```bash
+npm install
+make
+```
+
+Navigate to the *translate-all* directory containing this README and install
+```bash
+cd examples/translate-all
 npm install
 ```
 
--- a/examples/translate-all/src/helpers.js
+++ b/examples/translate-all/src/helpers.js
@@ -1,0 +1,22 @@
+export function arrayLocalizableToTextLocalizables (
+  {locale, srcLocale, srcValue, localize}
+) {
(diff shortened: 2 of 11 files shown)


geigerzaehler:
Could you add this widget (with a GIF) to the main README?

examples/translate-all/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+First set the access token on your environment, then create the widget on Contentful:
+```bash
+export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=<contentfulManagementApiToken>
+contentful-widget create --space-id <yourSpaceId>
+```
+
+Serve on `http://localhost:3000`:
+```bash
+gulp watch
```
geigerzaehler: You need to built the widget api first, that is run `make` in the repo’s main dir.

examples/translate-all/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+    translateButton.disabled = false
+  })
+
+  translateButton.addEventListener('click', () => {
+    translateButton.classList.add('cf-is-loading')
+    translateButton.disabled = true
+
+    newLocalizer().localize(localesGenerator.generateLocalizables())
+      .then(releaseButton)
+      .catch(releaseButton)
```
geigerzaehler: If the promise implementation supports it we should use `.finally(releaseButton)`

examples/translate-all/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+        cfApi.entry, fieldTypes, cfApi.locales.default)
+
+    translateButton.disabled = false
+  })
+
+  translateButton.addEventListener('click', () => {
+    translateButton.classList.add('cf-is-loading')
+    translateButton.disabled = true
+
+    newLocalizer().localize(localesGenerator.generateLocalizables())
```
geigerzaehler: Since the localize only exposes one public method there is no need to use a class. It should just be a simple function.

examples/translate-all/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+    function releaseButton () {
+      translateButton.classList.remove('cf-is-loading')
+      translateButton.disabled = false
+    }
+  })
+
+  function newLocalizer () {
+    const copyUntranslatableLocales = $('#trall-copy-untranslatable').checked
+    return new Localizer({copyUntranslatableLocales})
```
geigerzaehler: No need to pass an object if this is the only option.

examples/translate-all/src/localizer.js
```diff
@@ -0,0 +1,63 @@
...
+    const {srcValue, valueType} = localizable
+    if (
+      ['Symbol', 'Text'].indexOf(valueType) > -1 &&
+      typeof srcValue === 'string'
+    ) {
+      return this._translateLocale(localizable)
+    } else if (
+      valueType === 'Array' && Array.isArray(srcValue)
+    ) {
+      const textLocalizables = arrayLocalizableToTextLocalizables(localizable)
```
geigerzaehler: I think it is the responsibility of the `LocalizableGenerator` to call this method.

examples/translate-all/src/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+) {
+  const localizedValues = srcValue.slice()
+  return srcValue.map((value, i) => {
+    return {
+      locale,
+      srcLocale,
+      srcValue: value,
+      valueType: 'Text',
+      localize: (newValue) => {
+        console.log('srcValue was', srcValue, 'set', i, 'new value is', localizedValues)
```
geigerzaehler: Please remove this.

examples/translate-all/src/localizer.js
```diff
@@ -0,0 +1,63 @@
+import yandexTranslator from './yandex-translator'
+import {arrayLocalizableToTextLocalizables, localeToLanguage} from './helpers.js'
```
geigerzaehler: We have this tiny function in a helper file if it is only used here?

DanweDE:
> Could you add this widget (with a GIF) to the main README?

Sure, but for the GIF we'll have to solve [TP#9448](https://contentful.tpondemand.com/entity/9448) first or I'd have to do an app reload which wouldn't look that nice in the GIF.

examples/translate-all/widget.json
```diff
@@ -3,5 +3,5 @@
   "name": "Translate All Content",
   "src": "http://localhost:3000/",
   "sidebar": true,
-  "fieldTypes": ["Symbol", "Text"]
+  "fieldTypes": ["Symbol", "Text", "Object"]
```
DanweDE: For sidebar widgets it doesn't actually make sense to restrict the field types at all, does it? But then, it isn't intuitive at all how we enable sidebar widgets in the CMApp to be used in a CT in the first place.

hlabas:
Hi,

The examples repository is now https://github.com/contentful/extensions.
Would you mind moving your PR there?
Thanks!

PR #77, opened 2016-01-20
Add an example for editing a linked entry inline

lynnagara:
Adds an example widget that allows user to edit a linked entry inline from the parent entry.


floelhoeffel:
Nice! How about adding functionality for 1..N links. I have a couple nice use cases in mind where that would be super useful. For example a simple label CT which holds translated labels and then a configuration CT which points to N labels. Inline editing the labels when editing the config would be much easier.

hlabas:
Hi,
Would you mind putting this example proposal to the new repository?
https://github.com/contentful/extensions
Thanks!

PR #93, opened 2016-04-25
Alloy-editor widget: Fix broken rich text editor

zeusdeux:


examples/alloy-editor/app.js
```diff
@@ -1,11 +1,14 @@
-import widget from '../../lib/api'
+import { init } from '../../lib/api'
+
+const AlloyEditor = window.AlloyEditor
+
+init((api) => {
```
geigerzaehler: Let’s call the parameter `widget` to be consistent with the rest of the examples and docs

examples/alloy-editor/README.md
```diff
@@ -6,7 +6,7 @@ This widgets uses the [Alloy Text Editor][alloy] to edit text fields.
 * Install dependencies with `npm install`
 * Fill in your authentication data and space id in the `Makefile`.
 * Run `make create` to upload your widget to the server
-* Run `make serve` to serve the widget
```
geigerzaehler: Let’s keep this. It is more verbose

examples/alloy-editor/app.js
```diff
@@ -1,11 +1,14 @@
-import widget from '../../lib/api'
+import { init } from '../../lib/api'
+
+const AlloyEditor = window.AlloyEditor
```
geigerzaehler: Please always use `let` for variable assignment. `const` is confusing since it still allows object mutation.

zeusdeux: I would rather keep `const` since I want the program to throw if someone reassigns `AlloyEditor`.

examples/alloy-editor/app.js
```diff
@@ -1,25 +1,29 @@
...
-  api.window.startAutoResizer()
-  let editor = AlloyEditor.editable('content')._editor
+const AlloyEditor = window.AlloyEditor
+
+init((widget) => {
   let currentValue
-  editor.setData(api.field.getValue())
 
-  api.field.onValueChanged((value) => {
+  widget.window.updateHeight(200)
```
DanweDE: Why set the height explicitly if in the next line you use the auto resize option?

geigerzaehler: Yes please

PR #95, opened 2016-05-18
Slug widget: Fix broken slug editor

zeusdeux:
Fix:

- Wrong method names
- Race conditions due to api not being loaded before the widget entry point was called
- How duplicates were looked up (this wasn't working at all)

Refactor:

- Rewrite using ES2015 to keep it consistent with other widgets
- Abstract away change handling into a function
- Rename functions based on what they do
- Added comments

### DIFF
--- a/examples/slug/Makefile
+++ b/examples/slug/Makefile
@@ -1,21 +1,19 @@
-# Put your token and space id here.
-TOKEN=f00b4r
-SPACE=my5p4c3
+# Uncomment the lines below after adding your data
+# export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=<your token>
+# export SPACE=<id of space you want to install the widget for>
 
-export PATH := ../../bin:${PATH}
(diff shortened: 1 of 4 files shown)


DanweDE:
Would be nice to have a quick summary in the commit msg what's been broken.

zeusdeux:
@danweDE added the changes in the PR description. Will that suffice?

DanweDE:
@zeusdeux Very nice elaborate summary but the best place for it would be the commit message so it's visible when browsing the `git log`.

zeusdeux:
@DanweDE done!

DanweDE:
> Rewrite using ES2015 to keep it consistent with other widgets

nice, but keep in mind that we chose to implement all widgets a little differently to show diverse ways of doing things. So consistency was never a requirement here.

zeusdeux:
@DanweDE I was talking about this with @floelhoeffel and he suggested that we show one solid way of doing stuff rather than many different approaches so that people can get up and running quicker. 
That is also the reason why I am moving our builds to browserify + babelify in most places. Converging it all into an approach that people can follow out of the box.

PR #97, opened 2016-05-26
Converge internal and public ui-extensions api

zeusdeux:
- [x] Expose `contentType`
- [x] Add private `setString` method
- [x] Change `setValue` to call `setString` or `setValue` based on type of widget
- [x] Update `signal` implementation to be in line with what's in UI repo
- [x] Expose `space.getPublishedEntries`
- [x] Expose `space.getPublishedAssets`
- [x] Fix broken `api.field.onValueChanged` behaviour (blocked till integration tests work so that we can test this breaking change properly)
- [x] Expose `field.removeValue`
- [x] Expose `field.onIsDisabledChanged`
- [x] Expose `field.setActive`
- [x] Expose `field.setInvalid`

### DIFF
--- a/lib/api/channel.js
+++ b/lib/api/channel.js
@@ -1,20 +1,22 @@
 import Promise from 'yaku'
-import Signal from './signal'
+import { Signal } from './signal'
 
 export default class Channel {
 
   constructor (sourceId, targetWindow) {
     this.sourceId = sourceId
-    this.targetWindow = targetWindow
+    this.targetWindow = targetWindow // contentful webapp window
     this._messageCount = 0
     this._messageHandlers = {}
     this._responseHandlers = {}
 
+    // window refers to iframe contentWindow
     window.addEventListener('message', (event) => {
       this._handleMessage(event.data)
     })
   }
 
+  // call method with name `method` exposed by contentful web app `window`
   call (method, ...params) {
     const messageId = this._send(method, params)
     return new Promise((resolve, reject) => {
--- a/lib/api/entry.js
+++ b/lib/api/entry.js
@@ -1,10 +1,10 @@
-import Signal from './signal'
+import { MemoizedSignal } from './signal'
 import Field from './field'
 
 export default
 function createEntry (channel, entryData, fieldInfo, defaultLocale) {
   let sys = entryData.sys
-  let sysChanged = new Signal()
+  let sysChanged = new MemoizedSignal(sys)
 
   channel.addHandler('sysChanged', (_sys) => {
     sys = _sys
--- a/lib/api/field-locale.js
+++ b/lib/api/field-locale.js
@@ -1,19 +1,27 @@
-import Signal from './signal'
+import { MemoizedSignal } from './signal'
 
 export default class FieldLocale {
 
-  constructor (channel, {id, locale, value}) {
+  constructor (channel, {id, locale, value, type, isDisabled}) {
     this.id = id
     this.locale = locale
+    this.type = type
     this._value = value
-    this._valueSignal = new Signal()
+    this._valueSignal = new MemoizedSignal(value)
+    this._isDisabledSignal = new MemoizedSignal(isDisabled)
     this._channel = channel
 
     channel.addHandler('valueChanged', (id, locale, value) => {
-      if (id === this.id && (!locale || locale === this.locale)) {
+      if (id === this.id &&
+          (!locale || locale === this.locale) &&
+          this._value !== value) {
         this._value = value
         this._valueSignal.dispatch(value)
       }
+    })
+
+    channel.addHandler('isDisabledChanged', isDisabled => {
+      this._isDisabledSignal.dispatch(isDisabled)
     })
   }
 
@@ -26,8 +34,25 @@ export default class FieldLocale {
     return this._channel.call('setValue', this.id, this.locale, value)
   }
 
+  removeValue () {
+    this._value = undefined
+    return this._channel.call('removeValue', this.id, this.locale)
+  }
+
+  setInvalid (isInvalid) {
+    return this._channel.call('setInvalid', isInvalid, this.locale)
+  }
+
+  setActive (isActive) {
+    return this._channel.call('setActive', isActive)
+  }
+
   onValueChanged (handler) {
     return this._valueSignal.attach(handler)
+  }
+
+  onIsDisabledChanged (handler) {
+    return this._isDisabledSignal.attach(handler)
   }
 
 }
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -6,12 +6,13 @@ import createSpace from './space'
 
 export var init = initializeApi(createWidgetAPI)
 
-function createWidgetAPI (channel, {entry, locales, field, fieldInfo}) {
+function createWidgetAPI (channel, {entry, locales, field, fieldInfo, contentType}) {
   return {
     locales,
     field: new FieldLocale(channel, field),
     entry: createEntry(channel, entry, fieldInfo, locales.default),
     space: createSpace(channel),
-    window: createWindow(channel)
+    window: createWindow(channel),
+    contentType: contentType
   }
 }
--- a/lib/api/signal.js
+++ b/lib/api/signal.js
@@ -1,4 +1,6 @@
-export default class Signal {
+require('core-js/es6/symbol')
+
+export class Signal {
   constructor () {
     this._id = 0
     this._listeners = {}
@@ -14,11 +16,38 @@ export default class Signal {
     if (typeof listener !== 'function') {
       throw new Error('listener function expected')
     }
-    var id = this._id++
+    const id = this._id++
     this._listeners[id] = listener
-    var self = this
-    return function removeListener () {
-      delete self._listeners[id]
+    // return function that'll detach the listener
+    return () => delete this._listeners[id]
+  }
+}
+
+const memArgsSymbol = Symbol('Private memoized arguments')
+
+export class MemoizedSignal extends Signal {
+  constructor (...memoizedArgs) {
+    if (!memoizedArgs.length) {
+      throw new Error('Initial value to be memoized expected')
     }
+    super()
+    this[memArgsSymbol] = memoizedArgs
+  }
+
+  dispatch (...args) {
+    this[memArgsSymbol] = args
+    super.dispatch(...args)
+  }
+
+  attach (listener) {
+    /*
+     * attaching first so that we throw a sensible
+     * error if listener is not a function without
+     * duplication of is function check
+     */
+    const detachListener = super.attach(listener)
+
+    listener(...this[memArgsSymbol])
+    return detachListener
   }
 }
--- a/test/unit/signal.spec.js
+++ b/test/unit/signal.spec.js
@@ -1,107 +1,141 @@
-import Signal from '../../lib/api/signal'
+import { MemoizedSignal, Signal } from '../../lib/api/signal'
 import {noop} from '../helpers'
 
-describe(`Signal`, () => {
-  describe(`constructor`, () => {
-    it(`creates a Signal instance`, () => {
-      expect(new Signal()).to.be.instanceof(Signal)
-    })
+describe('MemoizedSignal', () => {
+  it('calls the listener with the initial value', () => {
+    const spy = sinon.spy()
+
+    new MemoizedSignal('zweiundvierzig').attach(spy)
+    expect(spy).to.have.been.calledWithExactly('zweiundvierzig')
   })
 
-  describe(`instance`, () => {
+  it('throws if there is no initial value provided', () => {
+    expect(() => new MemoizedSignal()).to.throw()
+  })
+
+  it('calls listener with most recently dispatched value', () => {
+    const spy = sinon.spy()
+    const args = [1, 'string', null, {}]
+    const signal = new MemoizedSignal(42)
+
+    signal.dispatch(...args)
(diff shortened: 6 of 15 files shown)


lib/api/channel.js
```diff
@@ -32,15 +34,21 @@ export default class Channel {
     }
     return this._messageHandlers[method].attach(handler)
   }
-
+  /* eslint-disable max-len */
```
geigerzaehler: No. Rules are *not* meant to be broken.

lib/api/field-locale.js
```diff
@@ -1,18 +1,20 @@
...
     this._value = value
-    this._valueSignal = new Signal()
+    this._valueSignal = new MemoizedSignal(value)
     this._channel = channel
 
     channel.addHandler('valueChanged', (id, locale, value) => {
       if (id === this.id && (!locale || locale === this.locale)) {
-        this._value = value
-        this._valueSignal.dispatch(value)
+        if (this._value !== value) {
```
geigerzaehler: Why a nested `if`. Should all be part of one if. But please separate the conditions into variables.

lib/api/signal.js
```diff
@@ -1,24 +1,51 @@
...
+export class Signal {
   constructor () {
     this._id = 0
     this._listeners = {}
   }
 
   dispatch (...args) {
     for (let key in this._listeners) {
-      this._listeners[key](...args)
+      this._listeners[key].apply(this, args)
```
geigerzaehler: Why `this`? We don’t want to give listeners access to the signal instance.

test/unit/signal.spec.js
```diff
@@ -1,36 +1,167 @@
...
+import { MemoizedSignal, Signal } from '../../lib/api/signal'
 import {noop} from '../helpers'
 
-describe(`Signal`, () => {
-  describe(`constructor`, () => {
-    it(`creates a Signal instance`, () => {
+describe('MemoizedSignal', function () {
+  describe('constructor', () => {
+    it('creates a MemoizedSignal instance', () => {
+      expect(new MemoizedSignal(0)).to.be.instanceof(MemoizedSignal)
```
geigerzaehler: Kind of redundant

test/unit/signal.spec.js
```diff
@@ -1,36 +1,167 @@
...
+      })
+    })
+  })
+})
+
+
+describe('Signal', () => {
+  describe('constructor', () => {
+    it('creates a Signal instance', () => {
       expect(new Signal()).to.be.instanceof(Signal)
```
geigerzaehler: Also redundant

test/unit/signal.spec.js
```diff
@@ -1,36 +1,167 @@
-import Signal from '../../lib/api/signal'
+import { MemoizedSignal, Signal } from '../../lib/api/signal'
 import {noop} from '../helpers'
 
-describe(`Signal`, () => {
-  describe(`constructor`, () => {
-    it(`creates a Signal instance`, () => {
+describe('MemoizedSignal', function () {
```
geigerzaehler: Duplicates all the tests from `Signal`. We should either remove those or parametrize them over the constructor and run them for both classes.

test/unit/signal.spec.js
```diff
@@ -1,36 +1,167 @@
...
+  describe('constructor', () => {
+    it('creates a MemoizedSignal instance', () => {
+      expect(new MemoizedSignal(0)).to.be.instanceof(MemoizedSignal)
+    })
+  })
+
+  describe('instance', () => {
+    let memoizedSignal
+    let spies
```
geigerzaehler: Only needed by the `dispatch()` test, so please move them to that block

test/unit/signal.spec.js
```diff
@@ -1,36 +1,167 @@
...
+          for (let name in obj) {
+            expect(spies[name]).to.have.callCount(obj[name])
+          }
+        }
+
+        it('fires attached listeners in the same order they were attached', () => {
+          memoizedSignal.attach(spies.one)
+          memoizedSignal.attach(spies.two)
+          memoizedSignal.attach(spies.three)
+          memoizedSignal.dispatch()
```
geigerzaehler: Please do `spies.one.reset()` etc. So we make the correct assertion.

test/unit/signal.spec.js
```diff
@@ -1,36 +1,167 @@
...
+          sinon.assert.calledWithExactly(spies.one, initialValue)
+
+          memoizedSignal.dispatch(...args)
+          sinon.assert.calledWithExactly(spies.one, ...args)
+
+          memoizedSignal.dispatch()
+          sinon.assert.calledWithExactly(spies.one)
+        })
+
+        it('memoizes last dispatched value', () => {
```
geigerzaehler: Be mor descriptive: 'calls listener with most recently dispatched value'.

Please also add a test case for firing with the initial value provided to the constructor.

lib/api/entry.js
```diff
@@ -1,10 +1,10 @@
-import Signal from './signal'
+import { MemoizedSignal } from './signal'
 import Field from './field'
 
 export default
 function createEntry (channel, entryData, fieldInfo, defaultLocale) {
   let sys = entryData.sys
-  let sysChanged = new Signal()
+  let sysChanged = new MemoizedSignal(entryData.sys)
```
geigerzaehler: Tests are missing

lib/api/index.js
```diff
@@ -6,12 +6,13 @@ import createSpace from './space'
...
-function createWidgetAPI (channel, {entry, locales, field, fieldInfo}) {
+function createWidgetAPI (channel, {entry, locales, field, fieldInfo, contentType}) {
   return {
     locales,
     field: new FieldLocale(channel, field),
     entry: createEntry(channel, entry, fieldInfo, locales.default),
     space: createSpace(channel),
-    window: createWindow(channel)
+    window: createWindow(channel),
+    contentType: contentType
```
geigerzaehler: Tests are missing.

lib/api/field-locale.js
```diff
@@ -1,18 +1,20 @@
...
+import { MemoizedSignal } from './signal'
 
 export default class FieldLocale {
 
   constructor (channel, {id, locale, value}) {
     this.id = id
     this.locale = locale
     this._value = value
-    this._valueSignal = new Signal()
+    this._valueSignal = new MemoizedSignal(value)
```
geigerzaehler: Please add a test for triggering handlers with the initial value

geigerzaehler:
Please squash the commits

lib/api/channel.js
zeusdeux: This and following comments in the file are artefacts of me trying to understand the code. They will be removed before I open a clean PR. They are basically "notes" for my understanding.

test/unit/signal.spec.js
zeusdeux: It's been there since before which is why I didn't touch it.

test/unit/signal.spec.js
zeusdeux: Only exists since the spec had a test for this for old Signal implementation. Didn't want to remove stuff even though it seemed unnecessary to have it there.

test/unit/signal.spec.js
zeusdeux: 👍  for descriptive spec name
Test for firing with initial value is [here](https://github.com/contentful/widget-sdk/pull/97/files#diff-d26dedb39b54adbb3d03e9b2a41a3468R25).

lib/api/signal.js
```diff
@@ -1,24 +1,51 @@
...
+export class Signal {
   constructor () {
     this._id = 0
     this._listeners = {}
   }
 
   dispatch (...args) {
     for (let key in this._listeners) {
-      this._listeners[key](...args)
+      this._listeners[key].apply(null, args)
```
geigerzaehler: Please use rest parameters: `this._listeners[key](...args)`.

lib/api/signal.js
```diff
@@ -1,24 +1,51 @@
...
     this._listeners[id] = listener
-    var self = this
-    return function removeListener () {
-      delete self._listeners[id]
+    // return function that'll detach the listener
+    return () => delete this._listeners[id]
+  }
+}
+
+const memArgsSymbol = Symbol('Private memoized arguments')
```
geigerzaehler: [Not supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) by IE :(

zeusdeux: I was under the impression that `core-js` [has this polyfilled](https://github.com/zloirock/core-js/blob/master/es6/symbol.js).

test/unit/signal.spec.js
```diff
@@ -1,107 +1,152 @@
...
-describe(`Signal`, () => {
-  describe(`constructor`, () => {
-    it(`creates a Signal instance`, () => {
-      expect(new Signal()).to.be.instanceof(Signal)
-    })
-  })
 
-  describe(`instance`, () => {
-    let signal
+function runCommonSpecs (SignalConstructor, initialValue = 42) {
```
geigerzaehler: Please rename it to something meaningful like `behavesLikeSignal`.

Also the `initialValue` is never set as a prameter, so please remove it.

test/unit/signal.spec.js
```diff
@@ -1,107 +1,152 @@
...
-          spies = {
-            one: sinon.spy(),
-            two: sinon.spy(),
-            three: sinon.spy()
-          }
-        })
+    describe('dispatch()', () => {
+      it('returns nothing and does not fail', () => {
+        expect(signalInstance.dispatch()).to.equal(undefined)
+      })
```
geigerzaehler: We can get rid of one level of nesting if we remove this meaningless test

test/unit/signal.spec.js
```diff
@@ -1,107 +1,152 @@
...
       })
     })
   })
+}
+
+
+describe('MemoizedSignal', () => {
+  it('calls the listener with the initial value', () => {
+    const spy = sinon.spy()
+    const _ = new MemoizedSignal('zweiundvierzig').attach(spy)
```
geigerzaehler: remove the `const _ =` part.

test/unit/signal.spec.js
```diff
@@ -1,107 +1,152 @@
...
+  it('calls listener with most recently dispatched value', () => {
+    const spy1 = sinon.spy()
+    const spy2 = sinon.spy()
+    const args = [1, 'string', null, {}]
+    const signal = new MemoizedSignal(42)
+
+    signal.attach(spy1)
+    signal.dispatch(...args)
+    signal.attach(spy2)
+    sinon.assert.calledWithExactly(spy2, ...args)
```
geigerzaehler: We are not making assertions on `spy1`, so one spy suffices.

test/unit/signal.spec.js
```diff
@@ -1,107 +1,152 @@
...
 
-  describe(`instance`, () => {
-    let signal
+function runCommonSpecs (SignalConstructor, initialValue = 42) {
+  describe('instance', () => {
+    let signalInstance
+    let spies
     beforeEach(() => {
-      signal = new Signal()
+      signalInstance = new SignalConstructor(initialValue)
```
geigerzaehler: Why not call it `signal` instead of `signalInstance`.

test/unit/signal.spec.js
```diff
@@ -1,107 +1,152 @@
...
-            three: sinon.spy()
-          }
-        })
+    describe('dispatch()', () => {
+      it('returns nothing and does not fail', () => {
+        expect(signalInstance.dispatch()).to.equal(undefined)
+      })
 
+      describe('with previously attached listeners', () => {
         function expectCallCount (obj) {
```
geigerzaehler: Can we make this a method on spies. Otherwise it is not very clear that it depends on the state of spies.

test/unit/signal.spec.js
geigerzaehler: Also move it to the bottom. First show use, then definition.

lib/api/entry.js
geigerzaehler: :heavy_check_mark:

test/unit/signal.spec.js
zeusdeux: Gonna go with just `test`.
So `test(MemoizedSignal)` and `test(Signal)`

test/unit/signal.spec.js
zeusdeux: I am going to actually modify this to assert that dispatch _never_ throws. Imho, we would want to make sure that calling `dispatch` is always safe and won't blow up someone's widget.

test/unit/signal.spec.js
zeusdeux: Did that yesterday but changes we sitting locally. :)

test/unit/signal.spec.js
zeusdeux: 🆒

test/unit/signal.spec.js
```diff
@@ -1,107 +1,141 @@
...
+    sinon.assert.calledWithExactly(spy, ...args)
+  })
+
+  test(MemoizedSignal)
+})
+
+
+describe('Signal', () => test(Signal))
+
+function test (SignalConstructor, initialValue = 42) {
```
geigerzaehler: > Also the `initialValue` is never set as a prameter, so please remove it.

geigerzaehler:
Please add a changelog entry

zeusdeux:
Closing this in favour of #109

PR #100, opened 2016-05-30
Docs: Add screenshot for json form editor widget

zeusdeux:


DanweDE:
Why is there so much noise in the screenshot's path. Thomas' chessbord.gif path looks much nicer:
`http://contentful.github.io/widget-sdk/assets/chessboard.gif`

zeusdeux:
That is because I am using github cloud to store the images instead of polluting the repository itself.

DanweDE:
I think same holds true for Thomas' link. It does not reside in the repo itself. Just have a look.

zeusdeux:
@DanweDE Fixed for all docs PRs.

PR #111, opened 2016-06-14
Replace SlimerJS with PhantomJS

geigerzaehler:
Installation of SlimerJS package keeps failing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ lint:
 	eslint lib/ bin/ test/
 
 test:
-	xvfb-run karma start --single-run --browsers SlimerJS --reporters dots
+	karma start --single-run
 
 ci: test lint
 
--- a/test/unit/entry.spec.js
+++ b/test/unit/entry.spec.js
@@ -3,6 +3,7 @@ import {
   noop,
   describeAttachHandlerMember
 } from '../helpers'
+import {find} from 'lodash'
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -26,7 +26,7 @@ lint:
 	eslint lib/ bin/ test/
 
 test:
-	xvfb-run karma start --single-run --browsers SlimerJS --reporters dots
+	karma start --single-run
```
zeusdeux: 🙇

test/unit/entry.spec.js
```diff
@@ -50,7 +51,7 @@ describe(`createEntry()`, () => {
       })
       it(`got instantiated with its related constructor given field info`, () => {
         Object.getOwnPropertyNames(entry.fields).forEach((fieldId) => {
-          const info = fieldInfo.find((info) => info.id === fieldId)
+          const info = find(fieldInfo, (info) => info.id === fieldId)
```
zeusdeux: Why not stick with `Array.prototype.find`? A whole new dependency just to use `_.find` seems too much imho.

geigerzaehler: Because it is not supported by phantomjs :(

zeusdeux: Shouldn't our transpilation pipeline make that available?

PR #114, opened 2016-06-21
fix: Handle 'undefined' call result

geigerzaehler:
Some method calls only want to indicate completion by returning a promise that resolves to `undefined`. We now handle that case.


zeusdeux:
LGTM pending the lint error.

PR #125, opened 2016-06-29
Initial commit of the diff extension.

floelhoeffel:


examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+
+<!-- Contenful UI Extensions SDK API -->
+<script src="//contentful.github.io/ui-extensions-sdk/cf-widget-api.js"></script>
+<!-- Diff Lib -->
+<script src="//cdnjs.cloudflare.com/ajax/libs/jsdiff/2.2.2/diff.min.js"></script>
+
+<!-- Diffing UI Extension -->
+<script>
+
+  function DiffExtension(extensionsApi) {
```
zeusdeux: Capitalization makes this seem like a `class` that should be instantiated using `new`. `diffExtension` is better in my opinion as it's meant to be the callback passed to `init()`.

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+
+    extensionsApi.window.updateHeight();
```
zeusdeux: I would rather use `startAutoResizer()`

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+    // Current locale
+    var fieldLocale = extensionsApi.field.locale;
+    var fieldId = extensionsApi.field.id;
+
+    var publishedValue;
+
+    // Populate the input field with the value of the entry.field.
+    draftTextEl.value = extensionsApi.entry.fields[extensionsApi.field.id].getValue();
+
+    extensionsApi.field.onValueChanged(onValueChanged);
```
zeusdeux: Since this is meant to be an example, I would suggest you store the `detach` functions returned by this and `onSysChanged` and call 'em when this extension is unloaded. If you don't, the web app will try and dispatch a value to the handlers assigned here even after the extension is unloaded.

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+    extensionsApi.entry.onSysChanged(onSysChanged);
+    draftTextEl.addEventListener('input', keyboardInputHandler);
+
+    // Load published value from CDA.
+    loadPublishedValue(extensionsApi.entry.getSys().id)
```
zeusdeux: A `var entry = extensionsApi.entry` uptop would reduce noise from code imho :)

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Uses diff lib to render diff to DOM of extension.
+     */
+    function renderDiff() {
+
+      // Do not render if there is no diff.
+      if (publishedValue == draftTextEl.value) {
+        diffTextEl.innerHTML = '<i>Draft and published value are equal.</i>';
+        //diffTextEl.parentNode.style.display = 'none';
```
zeusdeux: Please remove if not required.

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+        return;
```
zeusdeux: This is implied and hence not required.

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+      var diff = JsDiff.diffWordsWithSpace(publishedValue, draftTextEl.value);
+
+      diff.forEach(function (part) {
+        var className = part.added ? 'added' : part.removed ? 'removed' : '';
+        var span = document.createElement('span');
+        span.className = className;
+        span.appendChild(document.createTextNode(part.value));
+        diffTextEl.appendChild(span);
+      });
+      extensionsApi.window.updateHeight();
```
zeusdeux: Is this necessary?

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+      return extensionsApi.space.getPublishedEntries({'sys.id': entryId})
+    }
+
+    /**
+     * Uses diff lib to render diff to DOM of extension.
+     */
+    function renderDiff() {
+
+      // Do not render if there is no diff.
+      if (publishedValue == draftTextEl.value) {
```
zeusdeux: `===` please :)

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+    var fieldId = extensionsApi.field.id;
+
+    var publishedValue;
+
+    // Populate the input field with the value of the entry.field.
+    draftTextEl.value = extensionsApi.entry.fields[extensionsApi.field.id].getValue();
+
+    extensionsApi.field.onValueChanged(onValueChanged);
+    extensionsApi.entry.onSysChanged(onSysChanged);
+    draftTextEl.addEventListener('input', keyboardInputHandler);
```
zeusdeux: `keyboardInputHandler` might be called before the `loadPublishedValue` resolves. Which means `publishedValue` will be `undefined`. This will cause `JsDiff.diffWordsWithSpace(publishedValue, draftTextEl.value);` on line `104` to throw.

examples/diff/index.html
```diff
@@ -0,0 +1,141 @@
...
+    function onValueChanged(val) {
+      draftTextEl.value = val;
+    }
+
+    /**
+     * Calls the callback every time the value of the system object is changed
+     * @param sys The sys object of the entry.
+     */
+    function onSysChanged(sys) {
+      if (publishedVersion != sys.publishedVersion) {
```
zeusdeux: Strict inequality here please `!==`

examples/diff/index.html
floelhoeffel: nope - removed!

examples/diff/index.html
floelhoeffel: Good catch - I will deactivate the input until we have a published value.

floelhoeffel:
Thanks! I implemented the changes, please have another look.

examples/diff/index.html
```diff
@@ -36,43 +37,55 @@
...
     var fieldLocale = extensionsApi.field.locale;
     var fieldId = extensionsApi.field.id;
 
-    var publishedValue;
-
-    // Populate the input field with the value of the entry.field.
-    draftTextEl.value = extensionsApi.entry.fields[extensionsApi.field.id].getValue();
+    // Callbacks for changes on the field or the entry.sys object.
+    // Detach functions could be used to clean up the callbacks.
+    var detachValueChanged = field.onValueChanged(onValueChanged);
```
zeusdeux: You could invoke the detach methods when this iframe is unloaded. So something like: 
```js
window.addEventListener('beforeunload', function () {
  detachValueChanged();
  detachSysChanged();
});
```

examples/diff/index.html
```diff
@@ -89,28 +102,32 @@
...
       }
 
       // Clear diff DOM element.
       diffTextEl.innerHTML = '';
 
-      //diffTextEl.parentNode.style.display = 'block';
-
       var diff = JsDiff.diffWordsWithSpace(publishedValue, draftTextEl.value);
 
+      var className;
```
zeusdeux: I would move `var className` and `var spanEl` inside the function passed to `forEach` since they are only used in there.

examples/diff/index.html
```diff
@@ -119,22 +136,24 @@
...
-      if (publishedVersion != sys.publishedVersion) {
+      if (sys.publishedVersion && publishedVersion !== sys.publishedVersion) {
         publishedValue = draftTextEl.value;
         publishedVersion = sys.publishedVersion;
         renderDiff();
       }
     }
   }
-  window.contentfulWidget.init(DiffExtension)
+  window.contentfulWidget.init(diffExtension)
```
zeusdeux: `;` at eol

examples/diff/index.html
```diff
@@ -0,0 +1,160 @@
...
+    var publishedVersion = extensionsApi.entry.getSys().publishedVersion;
+    var publishedValue;
+
+    // Current locale
+    var fieldLocale = extensionsApi.field.locale;
+    var fieldId = extensionsApi.field.id;
+
+    // Callbacks for changes on the field or the entry.sys object.
+    // Detach functions could be used to clean up the callbacks.
+    var detachValueChanged = field.onValueChanged(onValueChanged);
```
zeusdeux: You could invoke the detach methods when this iframe is unloaded. So something like: 
```js
window.addEventListener('beforeunload', function () {
  detachValueChanged();
  detachSysChanged();
});
```

examples/diff/index.html
```diff
@@ -0,0 +1,160 @@
...
+        diffTextEl.innerHTML = '<i>Draft and published value are equal.</i>';
+        return;
+      }
+
+      // Clear diff DOM element.
+      diffTextEl.innerHTML = '';
+
+      var diff = JsDiff.diffWordsWithSpace(publishedValue, draftTextEl.value);
+
+      var className;
```
zeusdeux: I would move `var className` and `var spanEl` inside the function passed to `forEach` since they are only used in there.

examples/diff/index.html
```diff
@@ -0,0 +1,160 @@
...
+     */
+    function onSysChanged(sys) {
+      if (sys.publishedVersion && publishedVersion !== sys.publishedVersion) {
+        publishedValue = draftTextEl.value;
+        publishedVersion = sys.publishedVersion;
+        renderDiff();
+      }
+    }
+  }
+  window.contentfulWidget.init(diffExtension)
```
zeusdeux: `;` at eol

examples/diff/index.html
zeusdeux: I have outlined an approach [using `beforeunload` here](https://github.com/contentful/ui-extensions-sdk/pull/125/files#r69042740).

examples/diff/index.html
floelhoeffel: Uh neat - I did not know about that event - thanks!

examples/diff/index.html
```diff
@@ -83,20 +84,38 @@
...
       field.setValue(draftTextEl.value);
     }
 
+
+    /**
+     * Event handler triggered before the window unloads.
+     * @param event The event object.
+     */
+    function onBeforeUnloadHandler(event) {
+      window.removeEventListener('onbeforeunload', onBeforeUnloadHandler);
```
zeusdeux: Is this needed? I am curious since when the window is unloaded, it should logically release all the resources it had held and thus cleanup _all_ handlers for all events on the page right?

floelhoeffel: Maybe not but I would consider it good practice to clean up listeners. I will also remove the keyboard listener!

examples/diff/README.md
```diff
@@ -0,0 +1,68 @@
...
+```bash
+cd examples/diff
+```
+- Create a configuration file with your credentials for Contentful
+```bash
+touch .env
+echo "SPACE_ID={YOUR-SPACE-ID}" >> .env
+echo "CONTENTFUL_MANAGEMENT_ACCESS_TOKEN={YOUR-MANAGEMENT-TOKEN}" >> .env
+```
+and replace space ID, management token and port accordingly.
```
zeusdeux: `port` isn't mentioned anywhere

examples/diff/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "draftPublishedDiff",
+  "version": "1.0.0",
+  "description": "Diffing UI extension for Contentful's web app.",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "create": "contentful-widget create --space-id $SPACE_ID",
+    "update": "contentful-widget update --space-id $SPACE_ID --force",
+    "dev": "contentful-widget update --space-id $SPACE_ID --force --src 'http://localhost:8001/index.html'"
```
zeusdeux: The docs start a local server on `3030` and this sets the src to `localhost:8001` which would mean that out of the box the widget would 404 for people. Maybe use the same ports everywhere for consistency.

examples/diff/package.json
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "author": "frederik@contentful.com",
+  "license": "ISC",
```
zeusdeux: You're using MIT license above.

zeusdeux:
LGTM. Thanks :)

floelhoeffel:
Thanks for the review @zeusdeux !

PR #138, opened 2016-09-05
Remove examples from README

geigerzaehler:
They now live in the [`contentful/extensions`](https://github.com/contentful/extensions) repo (See https://github.com/contentful/extensions/pull/5).

Also moved the “Getting Started” section up to be more prominent


hlabas:
LGTM

PR #142, opened 2016-09-20
Only pack necessary files for NPM

geigerzaehler:
Keeps the package size small and prevents leaking unwanted files when publishing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,15 @@
   "bugs": {
     "url": "https://github.com/contentful/ui-extensions-sdk/issues"
   },
+  "files": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,6 +9,13 @@
   "bugs": {
     "url": "https://github.com/contentful/ui-extensions-sdk/issues"
   },
+  "files": [
+    "dist/cf-extension-api.js*",
```
Khaledgarbaya: out of curiosity, why do we have `*` for these two files

geigerzaehler: I removed the globs and included the files explicitly.

PR #143, opened 2016-09-20
Switch to unpkg CDN

geigerzaehler:
* Do not create distributions tagged with versions anymore
* Do not update the file distribution in `./docs`.
* We keep the old dist files on GH Pages for backwards compatibility
* Document using unpkg.com to get distributed files

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -2,6 +2,8 @@
 
 ### API documentation and changelog
 
+We use semantic versioning.
+
 A commit that changes the public API of the SDK must come with a changelog entry
 and an updated to the API documentation.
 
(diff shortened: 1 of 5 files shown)


docs/ui-extensions-sdk-frontend.md
```diff
@@ -22,29 +22,18 @@ communicate with the Contentful Management App.
 You will need to include the `contentful-extension-sdk` library in your HTML5 app like
 so:
 ~~~html
-<script src="https://contentful.github.io/ui-extensions-sdk/cf-extension-api.js"></script>
+<script src="https://unpkg.com/contentful-ui-extensions-sdk@2"></script>
```
Khaledgarbaya: how about having `@latest` here instead of `@major` this way we encourage to use the latest version and we don't have to update it each time

webpack.config.js
```diff
@@ -1,16 +1,9 @@
...
-var entry = {}
-entry['cf-extension-api'] = './lib/api/index.js'
-tags.forEach((tag) => {
-  entry[`tagged/cf-extension-api-v${tag}`] = './lib/api/index.js'
-})
-
 module.exports = {
-  entry: entry,
+  entry: {
+    'cf-extension-api': './lib/api/index.js'
```
Khaledgarbaya: This a lot better already, but Is there a reason that we use an object as an entry here since we have on single entry why not having `entry:'./lib/api/index.js'`

DEVELOPING.md
```diff
@@ -23,21 +25,24 @@ can build the styleguide with `make styleguide` which will create HTML files in
...
 1. Add new release to changelog by renaming “Upcoming” to “v1.2.3”.
 2. Update API documentation for added methods to include the version they were
    added in (e.g. _“Since 1.2.3”_)
 3. Update the version in `package.json`.
-4. [Update GH Pages distribution](#updating-gh-pages-distribution)
-5. Commit to master.
-6. Tag commit as `vX.Y.Z` and push tag to GitHub
-7. Publish new version of package on `npm`
+4. Commit to master.
+5. Tag commit as `vX.Y.Z` and push tag to GitHub
```
Khaledgarbaya: Why are we tagging manually here, we can use semantic-release-cli to do this, we have that implemented in [contentful.js](https://github.com/contentful/contenful.js) the only thing is we should use conventional log commit messages, this way we'll have even release notes automatically generated

docs/ui-extensions-sdk-frontend.md
hlabas: Problem with `@latest` is that introducing breaking changes would likely break existing UI extensions developed if they reference it directly. I'd prefer absolute version usage so that it's up to users to decide when and how they want to migrate.

DEVELOPING.md
geigerzaehler: I like the idea but it is out of scope for this PR.

webpack.config.js
geigerzaehler: We want the output file to be named `cf-extension-api.js`. It would be named `main.js` if we followed your approach. We could control this with the `output.filename` configuration but I think this is not he right place: If we want to add mulptile targets we would need to revert again.

hlabas:
LGTM as well

PR #144, opened 2016-10-06
`field.setValue()` triggers `onValueChanged()`

geigerzaehler:


test/unit/field-locale.spec.js
```diff
@@ -99,49 +102,42 @@ describe('FieldLocale', () => {
...
     })
 
-    describe('targeted at another field’s id', () => {
-      it('does not update the value', () => {
-        const oldValue = field.getValue()
-        valueChangedHandler(`${field.id}-other-id`, defaultLocale, newValue)
+    it('calls handler when value change is received', () => {
+      const spy = sinon.spy()
+      field.onValueChanged(spy)
+      spy.reset()
```
DanweDE: The thing is, if you initialize it and then don't use it in `onValueChanged` like I said, then you don't need to reset it since it just got initialized (same state).

Or am I missing something?

geigerzaehler: I’m confused :confused:. Let’s discuss this f2f on monday.

DanweDE: 👍

DanweDE: After looking at it again - what I suggested makes no sense indeed. Sorry.

PR #163, opened 2018-03-14
Fix anchor link to first section

osartun:
The link to the first section in the ToC read `#inclusion-into-your-project` while the headline itself read `Inclusion in your project` resulting in the actual name `inclusion-in-your-project` which is why the link in the ToC didn't work. I fixed it by adjusting the link and text in the ToC.


floelhoeffel:
Good catch, thanks!

PR #168, opened 2018-06-29
feature(Add default outlines to all interactive elements)

wichniowski:
For improved accessability all interactive elements should have outlines in their focused/active state


wichniowski:
https://contentful.atlassian.net/browse/AUTH-44

PR #175, opened 2018-09-21
Move dev deps to devdeps

Khaledgarbaya:
We should move dependencies like `webpack`, stylus to `devDEpendencies`.
En users should not get these when installing the library

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,18 +27,18 @@
     "url": "git+https://github.com/contentful/ui-extensions-sdk.git"
   },
   "dependencies": {
+    "object-assign": "^4.1.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -27,18 +27,18 @@
     "url": "git+https://github.com/contentful/ui-extensions-sdk.git"
   },
   "dependencies": {
-    "babel-loader": "^6.4.1",
-    "babel-core": "^6.25.0",
-    "babel-preset-es2015": "^6.0.15",
-    "core-js": "^2.4.0",
```
backflip: Since this is referenced in `signal.js` I would keep it in `dependencies`.

PR #176, opened 2018-09-21
Moved some dependencies into devDependencies

backflip:
Until now, many actual dev dependencies were located in `dependencies`. While this is also not ideal with regards to the downloaded package's size, in our case it even lead to conflicts since the (pretty old) `webpack` and `babel-loader` packages were "overwriting" newer versions from other dependencies (I don't yet understand why npm would do this, to be honest).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,19 +27,15 @@
     "url": "git+https://github.com/contentful/ui-extensions-sdk.git"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


backflip:
Argh, sorry, missed https://github.com/contentful/ui-extensions-sdk/pull/175

I kept `core-js` in the main dependencies, though, since it is referenced in the source somewhere.

PR #188, opened 2019-01-22
Add navigator methods

jelz:


lib/navigator.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function createNavigator (channel) {
+  return {
+    openEntry: (id, opts) => {
+      return channel.call('navigate', { ...opts, entityType: 'Entry', id })
```
DanweDE: There could be many more use-cases for navigation than navigating to entities. E.g. what we are implementing right now is sliding-in bulk editor instances, so you could easily add a `openBulkEditor({ entryIds: [] })` or you could implement a command to navigate to the account view or entry list etc.

That makes me think that we should either have a less generic name for the command than "`navigate`" or we should have a more generic data structure easily supporting these other cases without creating a property hell.

PR #193, opened 2019-03-12
Add image upload methods

azer:
Adding the corresponding proxies of the image upload methods we've added to the web app

### DIFF
--- a/lib/space.js
+++ b/lib/space.js
@@ -32,12 +32,11 @@ const spaceMethods = [
   'unarchiveEntry',
   'unarchiveAsset',
 
-  'processAsset',
-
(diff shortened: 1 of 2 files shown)


lib/space.js
```diff
@@ -34,7 +34,10 @@ const spaceMethods = [
 
   'processAsset',
 
-  'getUsers'
+  'getUsers',
+
+  'createUpload',
```
jelz: 1. You could group these methods with `processAsset` as they belong to the same category
2. Add the new methods to the list in `test/unit/space.js`

azer:
@jelz comments addressed

PR #195, opened 2019-03-19
Added TS typings

suevalov:
* Better DX without Typescript
* Type-safety using Typescript

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -1,18 +1,19 @@
 declare module 'contentful-ui-extensions-sdk' {
 
-  interface Sys {
-    id: string;
-    type: string;
-  }
-
   interface SpaceMembership {
-    sys: Sys;
+    sys: {
+      id: string;
+      type: string;
+    };
     admin: boolean;
     roles: { name: string; description: string }[];
   }
 
   interface User {
-    sys: Sys;
+    sys: {
+      id: string;
+      type: string;
+    };
     firstName: string;
     lastName: string;
     email: string;
@@ -68,7 +69,7 @@ declare module 'contentful-ui-extensions-sdk' {
 
   /* Content Type API */
 
-  type ContentTypeField = {
+  interface ContentTypeField {
     disabled: boolean;
     id: string;
     localized: boolean;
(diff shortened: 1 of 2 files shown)


typings.d.ts
```diff
@@ -0,0 +1,308 @@
...
+      };
+      createdAt?: string;
+      createdBy?: {
+        sys: Sys;
+      };
+      updatedAt?: string;
+      updatedBy?: {
+        sys: Sys
+      };
+      linkType?: string;
```
jelz: This shouldn't be here

typings.d.ts
```diff
@@ -0,0 +1,308 @@
+declare module 'contentful-ui-extensions-sdk' {
+
+  interface Sys {
+    id: string;
+    type: string;
+    linkType?: string;
```
jelz: there should be a separate interface `LinkSys` which includes `linkType`. `linkType` is only valid in the context of a link.

typings.d.ts
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  interface SpaceMembership {
+    sys: Sys;
+    admin: boolean;
+    roles: { name: string; description: string }[];
+  }
+
+  interface ContentTypeSys {
+    sys: {
```
jelz: Can we just define actual sys properties here and just use it in `ContentTypeAPI` instead of doing intersection?

typings.d.ts
```diff
@@ -0,0 +1,308 @@
...
+  interface ParametersAPI {
+    installation: Object;
+    instance: Object;
+    invocation?: Object;
+  }
+
+  export interface ExtensionSDK {
+    field?: FieldAPI;
+    entry: EntryAPI;
+    contentType: ContentTypeAPI;
```
jelz: Since it's not an API I'd just call the type `ContentType`

typings.d.ts
```diff
@@ -0,0 +1,308 @@
...
+    space: SpaceAPI;
+    locales: LocalesAPI;
+    user: UserAPI;
```
jelz: Since it's not an API but just an entity I'd call the type `User`

typings.d.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  export const init: (initCallback: InitCallback) => void;
+  export const locations: {
+    LOCATION_ENTRY_FIELD: string;
+    LOCATION_ENTRY_FIELD_SIDEBAR: string;
+    LOCATION_ENTRY_SIDEBAR: string;
+    LOCATION_DIALOG: string;
+  }
+
+}
```
jelz: nl missing

typings.d.ts
```diff
@@ -0,0 +1,298 @@
...
+
+  interface EntryAPI {
+    getSys: () => Object;
+    onSysChanged: (callback: (sys: Object) => void) => Function;
+    fields: { [key: string]: EntryFieldAPI }
+  }
+
+  /* Content Type API */
+
+  type ContentTypeField = {
```
jelz: why is it a type alias?

typings.d.ts
```diff
@@ -0,0 +1,298 @@
...
+    linkType?: string;
+    items?: Object;
+  }
+
+  interface ContentType {
+    sys: {
+      type: string;
+      id: string;
+      version?: number;
+      space?: {
```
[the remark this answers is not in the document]
suevalov: done

typings.d.ts
suevalov: used `interface`

PR #212, opened 2019-06-18
page extensions navigation

davidfateh:
# Purpose of PR

Adding new method for navigating page extensions. Will have to see if this is appropriate to use with something like React Router (or any general routing system).

## PR Checklist

- [x] Tests are passing


package.json
```diff
@@ -17,6 +17,7 @@
     "lint": "eslint lib/ test/",
     "lint:fix": "eslint --fix lib/ test/",
     "build": "parcel build lib/index.js --no-source-maps --out-file cf-extension-api.js --global contentfulExtension",
+    "build:debug": "parcel build lib/index.js --out-file cf-extension-api.js --global contentfulExtension",
```
davidfateh: This allows source maps which allows for placing inline debuggers in the iframe. Pretty useful.

lib/navigator.js
```diff
@@ -1,4 +1,4 @@
-module.exports = function createNavigator(channel) {
+module.exports = function createNavigator(channel, extensionId) {
```
davidfateh: Adding this param in changes the way the tests work a bit. Going to have to look at it a little more and see if this is appropriate. Feedback is definitely welcome on the approach

suevalov: Let's call this variable `currentExtensionId` to make it more clear.

lib/navigator.js
```diff
@@ -16,6 +16,9 @@ module.exports = function createNavigator(channel) {
     },
     openNewAsset: opts => {
       return channel.call('navigateToContentEntity', { ...opts, entityType: 'Asset', id: null })
+    },
+    openPageExtension: opts => {
+      return channel.call('navigateToPageExtension', { extensionId, ...opts })
```
suevalov: `extensionId: currentExtensionId`

typings.d.ts
```diff
@@ -426,6 +432,7 @@ declare module 'contentful-ui-extensions-sdk' {
     LOCATION_ENTRY_SIDEBAR: string;
     LOCATION_DIALOG: string;
     LOCATION_ENTRY_EDITOR: string;
+    LOCATION_PAGE: string;
```
suevalov: We need to create one more SDK type:

```
  export type PageExtensionSDK = BaseExtensionSDK & {
    /** A set of IDs actual for the extension */
    ids: Pick<IdsAPI, 'environment' | 'space' | 'extension' | 'user'>;
  }
```

and update `init` function declaration

davidfateh:
@jelz I've added some new tests for the new `openPageExtension` method.

https://github.com/contentful/ui-extensions-sdk/pull/212/commits/2aadc4a7ed558c7d505845ce88b789f0036e0db2

PR #219, opened 2019-07-09
Added sidebar cypress test

suevalov:
# Purpose of PR

Check that sidebar extension is rendered.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/test/cypress/integration/entry-editor-extension.spec.js
+++ b/test/cypress/integration/entry-editor-extension.spec.js
@@ -1,0 +1,34 @@
+import { entry } from '../utils/paths'
+
+const post = {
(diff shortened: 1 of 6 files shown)


test/cypress/integration/entry-editor-extension.spec.js
```diff
@@ -0,0 +1,34 @@
+import { entry } from '../utils/paths'
+
+const post = {
+  id: '5mwUiJB2kThfAG9ZnRNuNQ',
+  title: 'My post with a custom entry editor',
+  body: ''
```
davidfateh: Might be worth actually adding a `body` in as it may catch false-positives where something else renders with an empty string.

PR #221, opened 2019-07-10
Test dialogs.openExtension method

suevalov:
# Purpose of PR

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/test/cypress/integration/page-extension.spec.js
+++ b/test/cypress/integration/page-extension.spec.js
@@ -3,7 +3,10 @@ import { pageExtension } from '../utils/paths'
 const selectors = {
   pageIFrame: '[data-test-id="page-extension"] iframe',
   page: '[data-test-id="my-page-extension"]',
-  openNewPathButton: '[data-test-id="open-new-path-button"]'
+  openNewPathButton: '[data-test-id="open-new-path-button"]',
+  openDialogExtensionButton: '[data-test-id="open-dialog-extension-button"]',
+  modalIFrame: '[data-test-id="cf-ui-modal"] iframe',
+  dialogExtension: '[data-test-id="my-dialog-extension"]'
 }
 
 context('Page extension', () => {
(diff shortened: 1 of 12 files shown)


test/cypress/integration/page-extension.spec.js
```diff
@@ -12,17 +14,20 @@ context('Page extension', () => {
...
 
+  it('opens a page extension and tests navigating within the page', () => {
     cy.get('@extension')
       .find(selectors.openNewPathButton)
       .click()
 
     cy.url().should('include', 'test-extension/new')
   })
+
+  /* Reusable tests */
```
davidfateh: Is this comment necessary?

test/cypress/integration/reusable/open-dialog-extension-test.js
```diff
@@ -0,0 +1,33 @@
...
+  dialogSelector = '[data-test-id="my-dialog-extension"]'
+) {
+  cy.waitForIFrame()
+  cy.get(selector).captureIFrameAs('dialogExtension')
+  cy.get('@dialogExtension')
+    .find(dialogSelector)
+    .should('be.visible')
+}
+
+export function openDialogExtensionTest() {
```
davidfateh: Maybe this function can take an argument that will get added to the description in the `it` statement to at least give some more context of what was actually being tested? Might not be necessary since it will be in a `describe`.

test/cypress/integration/page-extension.spec.js
suevalov: @davidfateh not necessary, but I think it's a good separator between tests specific to a location and reusable ones.

test/cypress/integration/reusable/open-dialog-extension-test.js
suevalov: yeah, maybe it makes sense. I'd like to add more tests to collect problems and nuisances to make an informed refactoring.

PR #222, opened 2019-07-12
Add "platformAlpha" namespace

jelz:
# Purpose of PR

adds a new `platformAlpha` namespace with the following API:

- `sdk.platformAlpha.app.isInstalled(): boolean`
- `sdk.platformAlpha.app.getParameters(): null | object`
- `sdk.platformAlpha.app.getCurrentState(): null | object`
- `sdk.platformAlpha.app.onConfigure(handler): void`

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -1,0 +1,69 @@
+const HOOK_STAGE_PRE_INSTALL = 'preInstall'
+const HOOK_STAGE_POST_INSTALL = 'postInstall'
+
+const isObject = o => typeof o === 'object' && o !== null && !Array.isArray(o)
+
+const runHandler = handler => {
+  handler = typeof handler === 'function' ? handler : () => {}
+
+  let resultMaybePromise
+  try {
+    resultMaybePromise = handler()
+  } catch (err) {
+    resultMaybePromise = false
+  }
+
+  const isPromise = resultMaybePromise && typeof resultMaybePromise.then === 'function'
+  const resultPromise = isPromise ? resultMaybePromise : Promise.resolve(resultMaybePromise)
+
+  return resultPromise.then(
+    result => {
+      if (result instanceof Error || result === false) {
+        return false
+      } else if (!isObject(result)) {
+        return {}
+      } else {
+        return result
+      }
+    },
+    () => false
+  )
+}
+
+module.exports = function createApp(channel) {
+  const handlers = {
+    [HOOK_STAGE_PRE_INSTALL]: null,
+    [HOOK_STAGE_POST_INSTALL]: null
+  }
+
+  const setHandler = (stage, handler) => {
+    if (handlers[stage]) {
+      throw new Error('Cannot register a handler twice.')
+    } else if (typeof handler !== 'function') {
+      throw new Error('Handler must be a function.')
+    } else {
+      handlers[stage] = handler
+    }
+  }
+
+  channel.addHandler('appHook', ({ stage, installationRequestId }) => {
+    return runHandler(handlers[stage]).then(result => {
+      channel.send('appHookResult', { stage, installationRequestId, result })
+    })
+  })
+
+  return {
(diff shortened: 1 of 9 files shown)


lib/app.js
```diff
@@ -0,0 +1,69 @@
+const HOOK_STAGE_PRE_INSTALL = 'preInstall'
+const HOOK_STAGE_POST_INSTALL = 'postInstall'
+
+const isObject = o => typeof o === 'object' && o !== null && !Array.isArray(o)
+
+const runHandler = handler => {
+  handler = typeof handler === 'function' ? handler : () => {}
```
davidfateh: Could short circuit everything below.

```suggestion
if (typeof handler !== 'function) {
  return Promise.resolve(false);
}
```

lib/app.js
```diff
@@ -0,0 +1,69 @@
...
+
+  let resultMaybePromise
+  try {
+    resultMaybePromise = handler()
+  } catch (err) {
+    resultMaybePromise = false
```
davidfateh: Could short circuit again here:
```suggestion
    return Promise.resolve(false);
```

lib/app.js
```diff
@@ -0,0 +1,69 @@
...
+  const handlers = {
+    [HOOK_STAGE_PRE_INSTALL]: null,
+    [HOOK_STAGE_POST_INSTALL]: null
+  }
+
+  const setHandler = (stage, handler) => {
+    if (handlers[stage]) {
+      throw new Error('Cannot register a handler twice.')
+    } else if (typeof handler !== 'function') {
+      throw new Error('Handler must be a function.')
```
davidfateh: It looks like you `throw` here if the handler isn't a function but you still check for it in the `runHandler` function but don't throw there. So that code path *seems* like it will never run anyway with this code in the way here.

lib/app.js
```diff
@@ -0,0 +1,69 @@
+const HOOK_STAGE_PRE_INSTALL = 'preInstall'
+const HOOK_STAGE_POST_INSTALL = 'postInstall'
+
+const isObject = o => typeof o === 'object' && o !== null && !Array.isArray(o)
+
+const runHandler = handler => {
```
davidfateh: Any reason not to use `async` syntax?

jelz: We don't use `async` in `lib/` and it means we don't have to package `regenerator-runtime`. I don't feel like doing it just to use `async` in this one file. Converting the whole project would be a different story ;)

lib/app.js
jelz: This method is used by a user to inject a handler.

This condition protects us against `app.onConfigure('boom')` calls from end-user.

`runHandler` checks if a handler is registered and provides a default if not b/c as per spec it's not required (no handler registered means no parameters will be used for an app which sometimes may be valid).

lib/app.js
jelz: done

lib/app.js
jelz: done

lib/app.js
```diff
@@ -0,0 +1,69 @@
...
+    result => {
+      if (result instanceof Error || result === false) {
+        return false
+      } else if (!isObject(result)) {
+        return {}
+      } else {
+        return result
+      }
+    },
+    () => false
```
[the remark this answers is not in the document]
jelz: done

lib/app.js
davidfateh: Yep, makes sense.

jelz:
@davidfateh missed your question

> Question: Are we just going to remove the platformAlpha from the sdk when we are ready to go? What's the benefit of having that right now?

We'll eventually move `sdk.platformAlpha.app` to `sdk.app`, around the time we do platform beta. I prefix it with this special namespace to indicate that the API offered is subject to change

PR #232, opened 2019-08-07
Added test on sdk.ids static methods values

MarynaG:
Added test, which verifies, that sdk.ids static methods have expected values

### DIFF
--- a/test/cypress/integration/fixtures/ids-data.json
+++ b/test/cypress/integration/fixtures/ids-data.json
@@ -1,0 +1,9 @@
+{
+    "contentType": "postWithCustomEntryEditor",
+    "entry": "5mwUiJB2kThfAG9ZnRNuNQ",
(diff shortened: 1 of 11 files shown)


test/cypress/integration/fixtures/ids-data.json
```diff
@@ -0,0 +1,22 @@
...
+    "entryEditorExtension":
+    {
+        "entry": "5mwUiJB2kThfAG9ZnRNuNQ",
+        "contentType": "postWithCustomEntryEditor"
+    },
+    "fieldExtension":
+    {
+        "entry": "1MDrvtuLDk0PcxS5nCkugC",
+        "contentType": "post",
+        "field": "bdpb"
```
suevalov: It might be bug. Let me figure out why we receive this value.

PR #238, opened 2019-08-21
Added sdk.location.is method tests

MarynaG:
# Purpose of PR

Added tests, which verifies sdk.location.is method is true for all extension types.

### DIFF
--- a/test/cypress/integration/dialog-extension.spec.js
+++ b/test/cypress/integration/dialog-extension.spec.js
@@ -72,6 +72,10 @@ context('Dialog extension', () => {
   it('verifies sdk.location.is dialog', () => {
     cy.getSdk(iframeDialogSelector).then(sdk => {
       expect(sdk.location.is('dialog')).to.equal(true)
+      expect(sdk.location.is('entry-editor')).to.equal(false)
+      expect(sdk.location.is('entry-field')).to.equal(false)
+      expect(sdk.location.is('page')).to.equal(false)
+      expect(sdk.location.is('entry-sidebar')).to.equal(false)
     })
   })
 
--- a/test/cypress/integration/dialog-extension.spec.js
+++ b/test/cypress/integration/dialog-extension.spec.js
@@ -72,7 +72,6 @@ context('Dialog extension', () => {
 
   it('verifies sdk.location.is dialog', () => {
     cy.getSdk(iframeDialogSelector).then(sdk => {
-      expect(sdk.location.is('dialog')).to.equal(true)
       verifyLocation(sdk, 'dialog')
     })
   })
--- a/test/cypress/integration/dialog-extension.spec.js
+++ b/test/cypress/integration/dialog-extension.spec.js
@@ -69,6 +69,12 @@ context('Dialog extension', () => {
     })
   })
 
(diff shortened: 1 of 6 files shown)


test/cypress/integration/dialog-extension.spec.js
```diff
@@ -69,6 +69,12 @@ context('Dialog extension', () => {
     })
   })
 
+  it('verifies sdk.location.is dialog', () => {
+    cy.getSdk(iframeDialogSelector).then(sdk => {
+      expect(sdk.location.is('dialog')).to.equal(true)
```
jelz: let's test it's `false` for other values

davidfateh: Should your suggestion also apply to the other tests below?

jelz: https://github.com/contentful/ui-extensions-sdk/pull/238#pullrequestreview-278198669 ;)

MarynaG: done, thanks!

jelz:
What @davidfateh said in https://github.com/contentful/ui-extensions-sdk/pull/238#pullrequestreview-278246992

we should define a list of valid values:

```js
const VALID_LOCATIONS = ['dialog', 'entry-field', 'entry-sidebar' /* ... */];
```

and prepare an assertion function as proposed by David.

I'd also test for some input which is not on a list of valid locations:

```js
expect(sdk.location.is('BOOOOOM!')).to.equal(false)
```

test/cypress/utils/verify-location.js
```diff
@@ -0,0 +1,10 @@
+import locations from '../../../lib/locations'
+
+export function verifyLocation(sdk, location) {
+  expect(sdk.location.is(location)).to.equal(true)
+  let invalidLocations = Object.values(locations)
+  invalidLocations = invalidLocations.filter(item => item !== location)
+  invalidLocations.push('test-value')
+  console.log(invalidLocations)
```
davidfateh: Remove.
```suggestion

```

test/cypress/utils/verify-location.js
```diff
@@ -0,0 +1,10 @@
...
+  invalidLocations.every(item => expect(sdk.location.is(item)).to.equal(false))
```
davidfateh: Minor, but I think a `forEach` would make more sense here as we don't use the return value of this code.
```suggestion
  invalidLocations.forEach(item => expect(sdk.location.is(item)).to.equal(false))
```

test/cypress/utils/verify-location.js
```diff
@@ -0,0 +1,9 @@
+import locations from '../../../lib/locations'
+
+export function verifyLocation(sdk, location) {
+  expect(sdk.location.is(location)).to.equal(true)
+  let invalidLocations = Object.values(locations)
+  invalidLocations = invalidLocations.filter(item => item !== location)
+  invalidLocations.push('test-value')
```
jelz: Let's push `invalid-location` so it's more obvious

MarynaG: fixed.

PR #242, opened 2019-08-29
add test sdk parameters methods

MarynaG:
# Purpose of PR
Tests for installation, instance and invocation parameters have been added.

### DIFF
--- a/test/cypress/integration/dialog-extension.spec.js
+++ b/test/cypress/integration/dialog-extension.spec.js
@@ -14,7 +14,7 @@ import { verifyLocation } from '../utils/verify-location'
 import {
   verifySdkInstallationParameters,
   verifySdkInvocationParameters
-} from './reusable/open-sdk-parameters-test'
+} from '../utils/verify-parameters'
 
 const post = {
   id: '3MEimIRakHkmgmqvp1oIsM',
@@ -62,7 +62,7 @@ context('Dialog extension', () => {
     cy.get('@extension').should('be.visible')
   })
 
-  it('sdk.ids static methods have expected values', () => {
+  it('verifies sdk.ids static methods have expected values', () => {
     cy.getSdk(iframeDialogSelector).then(sdk => {
       expect(sdk.ids.contentType).to.equal(undefined)
       expect(sdk.ids.entry).to.equal(undefined)
@@ -80,7 +80,7 @@ context('Dialog extension', () => {
     })
   })
 
-  it('sdk.parameters have expected values', () => {
+  it('verifies sdk.parameters have expected values', () => {
     cy.getSdk(iframeDialogSelector).then(sdk => {
       verifySdkInstallationParameters(iframeDialogSelector)
       verifySdkInvocationParameters(iframeDialogSelector)
--- a/test/cypress/integration/dialog-extension.spec.js
+++ b/test/cypress/integration/dialog-extension.spec.js
@@ -11,6 +11,7 @@ import {
   openErrorNotificationTest
 } from './reusable/open-notifications-test'
 import { verifyLocation } from '../utils/verify-location'
+import { verifySdkInstallationParameters } from './reusable/open-sdk-parameters-test'
 
 const post = {
   id: '3MEimIRakHkmgmqvp1oIsM',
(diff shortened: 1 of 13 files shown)


test/cypress/integration/dialog-extension.spec.js
```diff
@@ -76,6 +80,13 @@ context('Dialog extension', () => {
     })
   })
 
+  it('sdk.parameters have expected values', () => {
+    cy.getSdk(iframeDialogSelector).then(sdk => {
```
davidfateh: Are these tests able to handle async code without any intervention? If not, these tests aren't actually running.
What I mean by this is the test would need to look something like this:

```js
it('sdk.parameters have expected values', async () => {
await   verifySdkInstallationParameters(iframeDialogSelector)
await   verifySdkInvocationParameters(iframeDialogSelector)
})
```

or

```js
it('sdk.parameters have expected values', (done) => {   
verifySdkInstallationParameters(iframeDialogSelector).then(done)
verifySdkInvocationParameters(iframeDialogSelector).then(done)
})
```

suevalov: @davidfateh https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Chains-of-Commands

> If you’re a modern JS programmer you might hear “asynchronous” and think: why can’t I just use async/await instead of learning some proprietary API? Cypress’s APIs are built very differently from what you’re likely used to.

test/cypress/integration/entry-editor-extension.spec.js
```diff
@@ -12,6 +12,7 @@ import {
   openErrorNotificationTest
 } from './reusable/open-notifications-test'
 import { verifyLocation } from '../utils/verify-location'
+import { openSdkParametersTest } from './reusable/open-sdk-parameters-test'
```
suevalov: I would rather split it to `verifySdkInstallationParameters` and `verifySdkInstanceParameters` to make it explicit what parameters are present.

test/cypress/integration/dialog-extension.spec.js
davidfateh: Thanks Alex, didn't know that!

davidfateh:
@suevalov What do you mean by page invocation parameters? A page extension should just use all the same parameters as any other location.

suevalov:
@davidfateh if you open page extension like this:
```
navigator.openPageExtension({path: '/settings/features'});
```
the `path` will be in `sdk.parameters.invocation`, right?

davidfateh:
@suevalov Yes, thanks for clarifying.

jelz:
I would also test default values: if there is a parameter defined with a default value but it's not provided then the SDK should return the default

MarynaG:
> I would also test default values: if there is a parameter defined with a default value but it's not provided then the SDK should return the default

Added test for non-default installation parameter.

@suevalov Could you take a look please and help me to identify, where instance parameters are set? Thank you!

suevalov:
@jelz @MarynaG I wouldn't block merging this PR, cause adding we need to add ability to set installation parameters to `create-contentful-extension` first.

suevalov:
Created an issue https://github.com/contentful/create-contentful-extension/issues/48

jelz:
@suevalov but we already pass installation parameters: https://github.com/contentful/ui-extensions-sdk/pull/242/files#diff-a67fdb72463fcfd2fbc2c601ba0d8e54R30

What is missing is setting instance parameters. To be hones I'm clueless - I've grepped the whole project and there's no single reference to editor interfaces (and this is where instance parameters are set). Do you know where do we do editor interface manipulation in the test code?

suevalov:
@jelz We don't manipulate with `editorInterface` in code. Everything is setup up in `master` environment manually.

Maybe we could set up `instance` params manually as well and tests that they override default params. WDYT?

jelz:
🤦‍♂ now I get it.

Yes, we should do it `master`

test/cypress/integration/field-extension.spec.js
```diff
@@ -73,6 +74,7 @@ context('Field extension', () => {
   it('verifies sdk.parameters have expected values', () => {
     cy.getSdk(iframeSelector).then(sdk => {
       verifySdkInstallationParameters(iframeSelector)
+      parameters.instance.instanceParameterEnumId = 'option2'
```
suevalov: a comment would help here.

MarynaG: done, thanks.

PR #246, opened 2019-10-08
adding setDirty and setPristine methods

davidfateh:
This will enable the SDK to contact the web app.

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -447,6 +447,10 @@ declare module 'contentful-ui-extensions-sdk' {
         getCurrentState: () => null | Object
         /** Registers a handler to be called to produce parameters and target state for an App **/
         onConfigure: (handler: Function) => void
+        /** Sets the state to be dirty and able to click the install/save button **/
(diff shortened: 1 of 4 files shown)


typings.d.ts
```diff
@@ -447,6 +447,10 @@ declare module 'contentful-ui-extensions-sdk' {
         getCurrentState: () => null | Object
         /** Registers a handler to be called to produce parameters and target state for an App **/
         onConfigure: (handler: Function) => void
+        /** Sets the state to be dirty and able to click the install/save button **/
+        setDirty: () => void
+        /** Sets the state to be pristine and disables the click to the install/save button **/
+        setPristine: () => void
```
jelz: Actually all methods but `onConfigure` are async (`return channel.call(...)` returns a promise). Can you tweak all the typings?

davidfateh: Updated

davidfateh:
@jelz thanks for taking a look.

PR #251, opened 2019-11-08
Making init() generic

fsubal:
# Purpose of PR

SDK's `init` often forces you to use `as` assertion, even when you know it would only be, for example, `FieldExtensionSDK`.

```ts
init(sdk => {
  const initialValue = (sdk as FieldExtensionSDK).field.getValue()

  render(<App sdk={sdk as FieldExtensionSDK} initialValue={initialValue} />, root)
})
```

I want to use generic to express it would only be a specific type of SDK.

```ts
init<FieldExtensionSDK>(sdk => {
  const initialValue = sdk.field.getValue()

  render(<App sdk={sdk} initialValue={initialValue} />, root)
})

// Or even...
init<FieldExtensionSDK | SidebarExtensionSDK>(sdk => { ... })
```

スクリーンショット 2019-11-08 10 23 04

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -455,7 +455,9 @@ declare module 'contentful-ui-extensions-sdk' {
     }
   }
 
-  export const init: (initCallback: (sdk: FieldExtensionSDK | SidebarExtensionSDK | DialogExtensionSDK | EditorExtensionSDK | PageExtensionSDK | AppExtensionSDK) => any) => void;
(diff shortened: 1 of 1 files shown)


jelz:
@suevalov is it fine or is it a breaking change?

suevalov:
@jelz Looks good to me too. It's not a breaking cause there's a default value for the generic present.

PR #263, opened 2020-01-15
"App" namespace

jelz:
# Purpose of PR

Introduce `app.sdk`. Patch bump for the transition period. Minor version bump will follow (for stabilizing the API fully).

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -261,7 +261,7 @@ declare module 'contentful-ui-extensions-sdk' {
     intent?: 'primary' | 'positive' | 'negative'
   }
 
-  interface OpenExtensionOptions {
+  interface OpenCustomWidgetOptions {
     id?: string
     width?: number | 'small' | 'medium' | 'large' | 'fullWidth'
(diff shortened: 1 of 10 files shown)


typings.d.ts
```diff
@@ -465,25 +468,27 @@ declare module 'contentful-ui-extensions-sdk' {
...
 
+  interface AppConfigAPI {
+    /** Tells the web app that the app is loaded */
+    setReady: () => Promise<void>
+    /** Returns true if an App is installed **/
+    isInstalled: () => Promise<boolean>
+    /** Returns parameters of an App, null otherwise **/
+    getParameters: () => Promise<null | Object>
+    /** Returns current state of an App, null otherwise **/
+    getCurrentState: () => Promise<null | Object>
```
davidfateh: I think it would be nice to allow the user to define the type instead of having to use `as`.
```suggestion
    getCurrentState<T>: () => Promise<null | T>
```

This comment applies to this and `getParameters` above.

PR #266, opened 2020-01-27
feat(locales): expose locales.optional and locales.direction

suevalov:
# Purpose of PR

Provide more useful information about locales that might be needed in some extensions.

## PR Checklist

- [x] Tests are passing


DanweDE:
Just an observation: We're having a somewhat odd mix of plural and singular words in `sdk.locales.` :thinking:

PR #270, opened 2020-02-12
Add entry field method overloads

taye:
# Purpose of PR

Overload entry field `onValueChanged` and `onIsDisabledChanged` methods.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.10.9",
+  "version": "3.10.10",
   "author": "Contentful GmbH",
   "license": "MIT",
   "main": "dist/cf-extension-api.js",
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -89,11 +89,6 @@ declare module 'contentful-ui-extensions-sdk' {
 
   /* Entry API */
 
-  type onValueChangedType = (callback: (value: any) => void) => Function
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.10.9",
+  "version": "3.10.10",
```
suevalov: Please, update version `package-lock.json` as well.

typings.d.ts
```diff
@@ -115,11 +110,15 @@ declare module 'contentful-ui-extensions-sdk' {
     /** Removes the value for the field and locale. */
     removeValue: (locale?: string) => Promise<void>
     /** Calls the callback every time the value of the field is changed by an external event or when setValue() is called. */
-    onValueChanged: onValueChangeWithLocaleType
-    onValueChanged: onValueChangedType
+    onValueChanged: {
```
suevalov: TIL for me! 👍

package.json
jelz: run `npm i` so package-lock gets updated with new version

typings.d.ts
```diff
@@ -115,11 +110,15 @@ declare module 'contentful-ui-extensions-sdk' {
...
+      (callback: (value: any) => void): Function
```
davidfateh: Is using `Function` ok? Should we simply instead do `() => void` and avoid using the Primitive?

taye: yeah `() => void` should be good

PR #280, opened 2020-03-26
fix: delete stale environments

FLoppix:
# Purpose of PR
Adding a script that checks if there are any stale environments from a previous integration test run that has not finished. Will check for environments that are older than `1 day` for now and remove them. Environments that are `master` or have the name `test` in it are not considered.

## PR Checklist

- [x] Tests are passing


test/integration/index.js
```diff
@@ -4,6 +4,7 @@ const asyncRetry = require('async-retry')
 const buildExtensions = require('./tasks/build-extensions')
 const deployExtensions = require('./tasks/deploy-extensions')
 const createEnvironment = require('./tasks/create-new-enviromenment')
+// const deleteStaleEnvironments = require('./tasks/delete-stale-environments')
```
davidfateh: Can this be removed?

test/integration/index.js
```diff
@@ -35,6 +36,7 @@ const cleanup = async () => {
         },
         { retries: 3 }
       )
+      // await deleteStaleEnvironments()
```
davidfateh: Removed?

suevalov: I'd place this command before creating an environment for the current test.

suevalov: And wrap it in `try catch`, we don't want to fail the whole integration script if any deletion fails.

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,40 @@
...
+  const isProtected = name => regexps.some(rgx => new RegExp(rgx).test(name))
+
+  const isStaleEnvironment = timeStamp => {
+    const environmentDate = new Date(timeStamp).getTime()
+    const difference = new Date().getTime() - environmentDate
+    return difference >= ONE_DAY_IN_MS
+  }
+  const deletedEnvironmentIds = []
+
+  items.forEach(async environment => {
```
suevalov: 1. I would run all deletions in parallel and wait for all of them to either finish or fail. 
2. If any of deletions fail - we don't care. I think and the race condition between different CI jobs is unavoidable so there will be some failed responses for sure.

FLoppix: good point ! Will add that

test/integration/index.js
FLoppix: No just wanted @suevalov to show where I could place that script and do not want run it unless it is approved

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,42 @@
...
+
+  await Promise.allSettled(
+    items.map(environment => {
+      const {
+        name,
+        sys: { createdAt, id }
+      } = environment
+      if (!isProtected(name) && isStaleEnvironment(createdAt)) {
+        try {
+          environment.delete()
```
suevalov: should we await this?

suevalov: I don't see there promise object is returned here

FLoppix: Sorry you were right! Was missing the `return` and `await` is not necessary since `Promise.allSettled` will handle everything until every `Promise` is resolved

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,43 @@
...
+module.exports = async (currentSpace = getCurrentSpace) => {
+  printStepTitle('Removing stale environments')
+
+  const space = await currentSpace()
+  const environments = await space.getEnvironments()
+  const { items } = environments
+
+  // filter for relevant environments
+  const regexps = [/^master$/g, /test/g]
+  const isProtected = name => regexps.some(rgx => new RegExp(rgx).test(name))
```
jelz: I'd avoid regexps for checks that can be expressed with code

```
const isProtected = name => name === 'master' || name.includes('test')
```

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,43 @@
...
+
+  const isStaleEnvironment = timeStamp => {
+    const environmentDate = new Date(timeStamp).getTime()
+    const difference = new Date().getTime() - environmentDate
```
jelz: `new Date().getTime()` => `Date.now()`

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,43 @@
+const getCurrentSpace = require('../contentful-client').getCurrentSpace
+const printStepTitle = require('../utils').printStepTitle
+const sleep = require('../utils').sleep
```
jelz: nit `const { printStepTitle, sleep } = require('../utils')`

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,43 @@
...
+    items.map(async environment => {
+      const {
+        name,
+        sys: { createdAt, id }
+      } = environment
+      if (!isProtected(name) && isStaleEnvironment(createdAt)) {
+        try {
+          deletedEnvironmentIds.push(id)
+          await sleep(200)
+          return environment.delete()
```
jelz: You need to do `return await environment.delete()` for async errors to be handled in `catch`

test/integration/tasks/delete-stale-environments.js
```diff
@@ -0,0 +1,43 @@
...
+  const deletedEnvironmentIds = []
+  await Promise.allSettled(
+    items.map(async environment => {
+      const {
+        name,
+        sys: { createdAt, id }
+      } = environment
+      if (!isProtected(name) && isStaleEnvironment(createdAt)) {
+        try {
+          deletedEnvironmentIds.push(id)
```
jelz: I'd expect the ID to be pushed once we're done with deletion.

I'd also consider collecting IDs of environments that couldn't be deleted and presenting a difference (all minus failed). Or presenting the failed ones

test/integration/tasks/delete-stale-environments.js
jelz: I'd actually run them in series... Reason: rate limits

FLoppix: Ya I learned it the hard way. Changed it to run them in a sequential order.

PR #283, opened 2020-04-09
remove guard in onConfigure* handlers

anho:
# Purpose of PR

Removing the guard in the `onConfigure*` handlers gives more flexibility to devs.

## PR Checklist

- [x] Tests are passing


davidfateh:
Looks like tests need updating.

PR #285, opened 2020-04-15
App page navigation

davidfateh:
# Purpose of PR
Adding in a new method to allow for navigation to app page locations.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/navigator.js
+++ b/lib/navigator.js
@@ -1,6 +1,6 @@
 const { Signal } = require('./signal')
 
-module.exports = function createNavigator(channel, currentExtensionId) {
+module.exports = function createNavigator(channel, ids) {
   const _onSlideInSignal = new Signal()
 
   channel.addHandler('navigateSlideIn', data => {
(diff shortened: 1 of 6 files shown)


lib/navigator.js
```diff
@@ -32,7 +32,10 @@ module.exports = function createNavigator(channel, currentExtensionId) {
       return channel.call('navigateToContentEntity', { ...opts, entityType: 'Asset', id: null })
     },
     openPageExtension: opts => {
-      return channel.call('navigateToPageExtension', { id: currentExtensionId, ...opts })
+      return channel.call('navigateToPage', { type: 'extension', id: ids.extension, ...opts })
+    },
+    openCurrentAppPage: opts => {
+      return channel.call('navigateToPage', { type: 'app', id: ids.app, ...opts })
```
jelz, line 38: I'd put `id` after `opts` so it's not possible to override

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.12.1",
+  "version": "3.13.0",
```
jelz, line 4: I'd have one more small thing before wee do 3.13 - can we do it later?

jelz: Also let's wait for #284 - we should be merging tomorrow. cc @taye

lib/navigator.js
davidfateh: I did this purposely so it is possible to override. Right now the web app won't accept the override but in the future we may want to accept certain overrides.

PR #291, opened 2020-04-22
refactor: convert project to typescript

taye:
### Purpose of PR

Convert JS source and test files to TS. Compile library to `dist/` and provide `typings.d.ts` as before.
### PR Checklist

- [x] Tests are passing


typings.d.ts
```diff
@@ -1,627 +1,25 @@
...
-  }
-
-  interface PageExtensionOptions {
-    /** If included, you can navigate to a different page extension. If omitted, you will navigate within the current extension. */
-    id?: string
-    /** Navigate to a path withing your page extension. */
-    path?: string
-  }
-
-  interface NavigatorSlideInfo {
```
suevalov, line 423: I think we used some of these types in `field-editors` even if was not exposed. How we make sure that it properly and doesn't introduce a breaking change?

lib/index.ts
```diff
@@ -0,0 +1,8 @@
+import createInitializer from './initialize'
+import createAPI from './api'
+import locations from './locations'
+
+export = {
```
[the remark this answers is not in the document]
taye: With a regular `export {}`, you'd have to do `require('contentful-ui-extensions-sdk').default` in non es6 projects. I've added a comment about this.

typings.d.ts
taye: It's still defined in `lib/types.ts` so internal files can import it from there.

suevalov: let's test it with field-editors before merging and releasing. ok?

lib/signal.ts
```diff
@@ -1,8 +1,6 @@
 class Signal {
-  constructor() {
-    this._id = 0
-    this._listeners = {}
-  }
+  _id = 0
+  _listeners = {}
```
shikaan, line 3: ⬆️

lib/signal.ts
```diff
@@ -24,11 +22,15 @@ class Signal {
 const memArgsSymbol = '__private__memoized__arguments__'
 
 class MemoizedSignal extends Signal {
+  [memArgsSymbol]: any[] = []
```
shikaan, line 25: This can also be made private

lib/space.ts
```diff
@@ -42,10 +42,13 @@ const spaceMethods = [
 
   'getAllScheduledActions',
   'getEntityScheduledActions'
-]
+] as const
 
-module.exports = function createSpace(channel, initialContentTypes) {
-  const space = {}
+export default function createSpace(channel, initialContentTypes) {
+  const space = {} as {
```
shikaan, line 48: Maybe we can define an interface here?

lib/window.ts
```diff
@@ -35,8 +35,8 @@ module.exports = function createWindow(currentWindow, channel) {
     currentWindow.removeEventListener('resize', autoUpdateHeight)
   }
 
-  function updateHeight(height) {
-    if (height == null) {
+  function updateHeight(height: number = null) {
```
shikaan, line 38:
```suggestion
  function updateHeight(height: number|null = null) {
```
I think now we're passing because of the permissive setup, but it won't pass later on

package.json
```diff
@@ -65,21 +69,20 @@
...
   },
   "lint-staged": {
-    "*.js": [
-      "prettier --write",
-      "eslint --fix",
-      "git add"
-    ],
     "*.ts": [
       "prettier --write",
+      "eslint --fix",
```
shikaan, line 85: Maybe we want to consider moving to tslint?

test/cypress/plugins/index.ts
```diff
@@ -1,5 +1,5 @@
-module.exports = (on, config) => {
-  on('before:browser:launch', (browser = {}, launchOptions) => {
+export default on => {
+  on('before:browser:launch', (browser = { name: '' }, launchOptions: { args: string[] }) => {
```
shikaan, line 2: Not sure, but we might have some interfaces available here from cypress itself

test/cypress/support/commands.ts
```diff
@@ -1,5 +1,17 @@
 import '@testing-library/cypress/add-commands'
 
+declare global {
```
shikaan, line 3: Why is this declaration like so? Is it because we need to reuse the interface across different files without importing it? In that case, it might be worth using a `d.ts`

lib/channel.ts
```diff
@@ -22,10 +20,11 @@ function waitForConnect(currentWindow, onConnect) {
 }
 
 class Channel {
-  constructor(sourceId, currentWindow) {
-    this._messageHandlers = {}
-    this._responseHandlers = {}
+  _messageHandlers = {}
+  _responseHandlers = {}
+  _send: ReturnType<typeof createSender>
```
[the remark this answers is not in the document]
jelz: Careful: we use some of these "private" outside of classes that define them

test/cypress/utils/verify-location.ts
```diff
@@ -2,8 +2,8 @@ import locations from '../../../lib/locations'
 
 export function verifyLocation(sdk, location) {
   expect(sdk.location.is(location)).to.equal(true)
-  let invalidLocations = Object.values(locations)
+  let invalidLocations = (Object as any).values(locations)
```
jelz, line 5: We're using ESNext lib so it should be just fine to do `Object.values`

test/cypress/support/commands.ts
taye: >Is it because we need to reuse the interface across different files without importing it?

Yes.

>In that case, it might be worth using a d.ts

Because it's a very small file with only a few definitions, I think it's more convenient to put it directly it in the definition file.

package.json
taye: Tslint is deprecated and they recommend using eslint with the typescript plugin

lib/channel.ts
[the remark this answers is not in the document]
taye: Changing them to `private` seems to work without type errors.

typings.d.ts
taye: It turns out that typescript was exposing all of these types even without `export` since they're inside the module declaration so I'll expose them again :+1:

lib/space.ts
taye: Nice catch. There was already a `SpaceAPI` interface for me to use :slightly_smiling_face:

test/cypress/plugins/index.ts
taye: Indeed we do

typings.d.ts
taye: With `lib/types.ts` being included in the package and referenced from `typings.d.ts`, typescript would suggest imports from `contentful-ui-extensions-sdk/lib/types`. To avoid this, I've added a `typings` npm script to generate a standalone `typings.d.ts` file in https://github.com/contentful/ui-extensions-sdk/pull/291/commits/cf21556ea7b68e7b0dd118f233128153e2be8242

PR #292, opened 2020-04-22
Inject custom API creator

jelz:
# Purpose of PR

Allow to create custom APIs. The creator function gets raw channel and init data and should return an object of methods.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/initialize.js
+++ b/lib/initialize.js
@@ -3,29 +3,40 @@ const { Promise } = require('es6-promise')
 const connect = require('./channel')
 
 module.exports = function createInitializer(currentWindow, apiCreator) {
-  const apiDeferred = createDeferred()
-  const channelDeferred = createDeferred()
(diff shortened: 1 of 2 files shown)


jelz:
That would be an undocumented option in the public SDK, yes.

This way we can prototype new methods as we build new apps instead of doing alpha namespaces etc.

davidfateh:
> That would be an undocumented option in the public SDK, yes.
> 
> This way we can prototype new methods as we build new apps instead of doing alpha namespaces etc.

So even if someone was to add their own function for that `makeCustomApi`, it would basically do nothing until we whitelist the ability on our end?

jelz:
> So even if someone was to add their own function for that makeCustomApi, it would basically do nothing until we whitelist the ability on our end?

Yes.

PR #296, opened 2020-04-28
feat: allow creation of field locale instance from field

Jwhiles:
# Purpose of PR
In order to use our Field Editor components with the Entry editor SDK, we need to have some way to get a `fieldLocale` object from an entry. I think the easiest way to do this is to make it possible to extract a fieldLocale from a field. 

Basically this PR just makes an existing private method public.

One thing that I'm not sure of, is how it's possible to determine from the Entry API which locale is currently being edited. 

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/lib/field.ts
+++ b/lib/field.ts
@@ -32,11 +32,11 @@ export default class Field {
   }
 
   getValue(locale) {
-    return this.getFieldLocale(locale).getValue()
+    return this._getFieldLocale(locale).getValue()
   }
 
   setValue(value, locale) {
-    return this.getFieldLocale(locale).setValue(value)
+    return this._getFieldLocale(locale).setValue(value)
   }
 
   removeValue(locale) {
@@ -48,7 +48,7 @@ export default class Field {
       handler = locale
       locale = undefined
     }
-    return this.getFieldLocale(locale).onValueChanged(handler)
+    return this._getFieldLocale(locale).onValueChanged(handler)
   }
 
   onIsDisabledChanged(locale, handler) {
@@ -57,12 +57,20 @@ export default class Field {
       locale = undefined
     }
 
-    return this.getFieldLocale(locale).onIsDisabledChanged(handler)
+    return this._getFieldLocale(locale).onIsDisabledChanged(handler)
   }
 
-  getFieldLocale(locale = this._defaultLocale) {
+  private _getFieldLocale(locale = this._defaultLocale) {
     assertHasLocale(this, locale)
     return this._fieldLocales[locale]
+  }
+
+  getLocaleSpecificField(locale) {
+    if (!locale) {
+      throw new Error(`getLocaleSpecificField must be passed a locale`)
+    }
+
+    return this._getFieldLocale(locale)
   }
 }
 
--- a/lib/field.ts
+++ b/lib/field.ts
@@ -32,11 +32,11 @@ export default class Field {
   }
 
   getValue(locale) {
-    return this._getFieldLocale(locale).getValue()
+    return this.getFieldLocale(locale).getValue()
   }
 
   setValue(value, locale) {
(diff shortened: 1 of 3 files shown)


kdamball:
LGTM but a related question, is there a reason this method was private before?

Jwhiles:
> LGTM but a related question, is there a reason this method was private before?

good question, and I'm honestly not really sure.

The method seems to come from [this commit](https://github.com/contentful/ui-extensions-sdk/commit/5a5ca05865b5b69e275035d78e5f0186953d54d5) - and it looks like it was introduced just to factor out some repetition.

lib/field.ts
```diff
@@ -57,11 +57,10 @@ export default class Field {
       locale = undefined
     }
 
-    return this._getFieldLocale(locale).onIsDisabledChanged(handler)
+    return this.getFieldLocale(locale).onIsDisabledChanged(handler)
   }
 
-  private _getFieldLocale(locale) {
-    locale = locale || this._defaultLocale
+  getFieldLocale(locale = this._defaultLocale) {
```
davidfateh, line 63: After reading the documentation PR you put up and looking at this method again, I'm a bit confused as a user. Why does a method called `getFieldLocale` have an optional argument to specify a locale?

Jwhiles: Yeah, the naming is not great. I think the distinction between a `field` and what we have called a `fieldLocale` is probably quite hard to understand for users full stop. I think it's made extra confusing by the fact that we don't really express the difference between them in the docs.

I'm really struggling to think of a good name for this - given the names of our types it should be something like `getFieldFromEntryField` - but that is kind of baffling

lib/field.ts
```diff
@@ -60,11 +60,18 @@ export default class Field {
     return this._getFieldLocale(locale).onIsDisabledChanged(handler)
   }
 
-  private _getFieldLocale(locale) {
-    locale = locale || this._defaultLocale
```
davidfateh, line 64: I think we should completely revert this change back to the way it was as the new logic is actually functionally different.

In this case if `locale` was falsy i.e. `''`, `0`,`undefined`, `null`, it would use the default.

In the new logic, using a default parameter only checks for `undefined`. Every other falsy value will NOT use the default.

The only caveat to this comment is if `assertHasLocale` would have done that for us. Either way it's probably safer to just not touch any of this code now that we are just introducing a new method completely.

lib/field.ts
```diff
@@ -60,11 +60,18 @@ export default class Field {
...
+  private _getFieldLocale(locale = this._defaultLocale) {
     assertHasLocale(this, locale)
     return this._fieldLocales[locale]
   }
+
+  getLocaleSpecificField(locale) {
+    if (!locale) {
+      throw new Error(`getLocaleSpecificField must be passed a locale`)
```
davidfateh, line 70: Nit:
```suggestion
      throw new Error('getLocaleSpecificField must be passed a locale')
```

lib/field.ts
```diff
@@ -65,6 +65,14 @@ export default class Field {
     assertHasLocale(this, locale)
     return this._fieldLocales[locale]
   }
+
+  getLocaleSpecificField(locale) {
```
jelz, line 69: I'd name it `getForLocale(localeCode)`;

you'll be calling it like this: `sdk.fields[id].getForLocale(localeCode)` - field is already mentioned

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.13.0",
+  "version": "3.14.0",
```
jelz, line 4: Let's do version bump separately.

PR #299, opened 2020-04-29
Add access API

jelz:
# Purpose of PR

Allow to check access level of a user.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -534,6 +534,23 @@ export interface SharedEditorSDK {
   }
 }
 
+export type CrudAction = 'create' | 'read' | 'update' | 'delete'
+export type PublishableAction = 'publish' | 'unpublish'
+export type ArchiveableAction = 'archive' | 'unarchive'
+
+export interface AccessAPI {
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -534,6 +534,23 @@ export interface SharedEditorSDK {
...
 
+export type CrudAction = 'create' | 'read' | 'update' | 'delete'
+export type PublishableAction = 'publish' | 'unpublish'
+export type ArchiveableAction = 'archive' | 'unarchive'
+
+export interface AccessAPI {
+  can(action: 'read' | 'update', entity: 'EditorInterface' | EditorInterface): Promise<boolean>
+  can<T = Object>(
+    action: CrudAction,
+    entity: 'ContentType' | ContentType | 'Asset' | 'Entry' | T
```
suevalov, line 545: What's the purpose of using ContentType (as a complex type) here?

jelz: With constraints you'll be able to express conditions based on properties of an entity, for example there can be a role allowing to create entities with a specific ID prefix. For that we need to pass an entity instead of a type. Type (string) checks are simple ones, contraint-based require an entity object.

PR #304, opened 2020-05-04
feat: listen to new schemaerrorschangedforfieldlocale events

Jwhiles:
# Purpose of PR
This change makes it possible to handle errors inside of entry extensions, by following a similar approach that was used to make `isDisabled` available in entry extensions.

This shouldn't be merged until we're confident that adding these events to user interface hasn't caused any issues.
## PR Checklist
- [x] Tests are passing

### DIFF
--- a/lib/field-locale.ts
+++ b/lib/field-locale.ts
@@ -44,8 +44,10 @@ export default class FieldLocale implements FieldAPI {
       }
     })
 
-    channel.addHandler('schemaErrorsChanged', errors => {
-      this._schemaErrorsChangedSignal.dispatch(errors)
+    channel.addHandler('schemaErrorsChangedForFieldLocale', (id, locale, errors) => {
+      if (id === this.id && locale === this.locale) {
+        this._schemaErrorsChangedSignal.dispatch(errors)
+      }
     })
   }
 
--- a/test/unit/field-locale.spec.ts
+++ b/test/unit/field-locale.spec.ts
@@ -112,7 +112,12 @@ describe('FieldLocale', () => {
   })
 
   describe('.onSchemaErrorsChanged(handler)', () => {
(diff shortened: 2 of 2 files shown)


test/unit/field-locale.spec.ts
```diff
@@ -112,7 +112,12 @@ describe('FieldLocale', () => {
   })
 
   describe('.onSchemaErrorsChanged(handler)', () => {
-    testChannelSignal('onSchemaErrorsChanged', 'schemaErrorsChanged')
+    testChannelSignal(
+      'onSchemaErrorsChanged',
+      'schemaErrorsChangedForFieldLocale',
+      [info.id, info.locale, [{ message: 'oh no' }]],
+      [[{ message: 'oh no' }]]
```
taye, line 119: webcomic-name-oh-no

lib/field-locale.ts
```diff
@@ -44,8 +44,10 @@ export default class FieldLocale implements FieldAPI {
       }
     })
 
-    channel.addHandler('schemaErrorsChanged', errors => {
-      this._schemaErrorsChangedSignal.dispatch(errors)
+    channel.addHandler('schemaErrorsChangedForFieldLocale', (id, locale, errors) => {
+      if (id === this.id && locale === this.locale) {
+        this._schemaErrorsChangedSignal.dispatch(errors)
```
jelz, line 49: This means the other `.dispatch()` call on `_schemaErrorsChangedSignal` can be removed. It'll stay as it is in older versions but from now on this handles all the fields.

Jwhiles: @jelz I think I've already replaced the previous dispatch logic, unless I'm missing something.

jelz: Ah sorry I thought it was an addition not replacement 🤦

Jwhiles:
I'm going to verify that these changes work with the field editors, and if so I will merge.

PR #305, opened 2020-05-05
fix: pass through required properties to fieldLocale from field

Jwhiles:
We actually need these info types now that we are able to get the fieldLocale instance
from the field. 😨 

## PR Checklist

- [ ] Tests are passing


Jwhiles:
> If these fields are required how come an error or warning isn't given?

@davidfateh this is actually one of things I mentioned in relation to why typing this project is hard - not 100% sure actually. I think our typescript setup is too lenient, and also the field.ts file doesn't currently have any typings!

PR #308, opened 2020-05-05
Updating the scheduled date typings to be strings

davidfateh:
This updates the typings for scheduled `ScheduledAction` to use a string for its date value.
I've included JSDocs to show up in most editors to indicate the format.

Screen Shot 2020-05-05 at 12 54 54

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -308,9 +308,11 @@ export type ScheduledAction = {
   sys: {
     id: string
     type: 'ScheduledAction'
-    createdAt: Date
+    /** ISO 8601 string */
+    createdAt: string
     createdBy: Link
-    canceledAt?: Date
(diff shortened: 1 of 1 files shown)


davidfateh:
> Does this count as a breaking change?

I'd say yes because it updates the typings but also since the typings were wrong anyone using the feature would have had to ignore the typings to not get yelled at by the `tsc`.

davidfateh:
> Thanks!
> 
> It's not a breaking change but a fix of wrong types in first place.

Technically it is but since I can't see anyone being able to use this feature without having to ignore the wrong type, it seems unlikely to 'break' for users as they were probably already ignoring the wrong type in the code. I'm going to include it in a round of updates that will be a patch in 13.14.**2**

PR #310, opened 2020-05-06
Add webhook methods

cezarsmpio:
# Purpose of PR

This PR mirrors 4 new methods, all related to webhook management:

- createWebhook
- createWebhookWithId
- getWebhook
- getWebhooks

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/lib/space.ts
+++ b/lib/space.ts
@@ -43,7 +43,12 @@ const spaceMethods: Array<keyof SpaceAPI> = [
   'getUsers',
 
   'getAllScheduledActions',
-  'getEntityScheduledActions'
+  'getEntityScheduledActions',
+  
+  'createWebhook',
+  'createWebhookWithId',
+  'getWebhook',
+  'getWebhooks'
 ]
 
 export default function createSpace(channel, initialContentTypes) {


lib/space.ts
```diff
@@ -43,7 +43,12 @@ const spaceMethods: Array<keyof SpaceAPI> = [
   'getUsers',
 
   'getAllScheduledActions',
-  'getEntityScheduledActions'
+  'getEntityScheduledActions',
+  
+  'createWebhook',
```
cezarsmpio, line 48: That's the one I need the most.

davidfateh:
I think we should have a quick discussion with more people to see if we should really add this for all our users. If so, I'd also suggest that you update the typings to help users using typescript understand what the arguments and return values are going to be for these new methods.

cezarsmpio:
Hey @davidfateh we have a case where we will need to add webhooks programmatically but we still have some ideas to tests and some decisions to make, I think regardless of those decisions, we will end up needing the webhook creation.

I will finish the PR, that's why I added the `blocked` label :) 

Thanks!

jelz:
I don't think we should be adding webhook methods, for a good reason. Please note it's a public repo, let's move discussion somewhere else @davidfateh @cezarsmpio

cezarsmpio:
Sure @jelz , just closed the PR! I can explain our needs and maybe you can give us better alternatives :)

PR #313, opened 2020-05-14
fix(test): flaky integration tests [EXT-1913]

FLoppix:
# Purpose of PR

Integration tests were flaky for quite some time. After some digging we found some issues that were fixed by this PR:

- Remove arbitrary `cy.waits` of 10 seconds 
- make use of cypress command retry-ability
- increase `defaultCommandTimeout` to 30 seconds

Also issues were caused by the web app not handling `429` with a proper retry logic resulting in errors that pages could not be rendered. https://github.com/contentful/user_interface/commit/879e269db738a9334e91537b20e6ad609f3764e3 Big shoutout to @jelz for finding this !

## PR Checklist

- [x] Tests are passing


FLoppix:
PS. Ignore the commit history. Hard times to figure this out :D 
@suevalov do we want to keep the videos or should we just go back to storing screenshots?

FLoppix:
failed test case will be fixed. Commited the wrong file for that

suevalov:
@FLoppix you're a hero. it a pain for sure.

Regarding video vs. screenshots I'm not sure, make it the way so it's easier for the team to figure out what's broken when it's broken :) 

Going forward, I'm pretty sure that we should mock all network interactions (contract tests)

PR #317, opened 2020-05-25
feat: add openCurrent method to dialogs api

FLoppix:
# Purpose of PR
In order to avoid confusion with extensions and apps this PR introduces a new method for the dialogs API to open the `current` extension or app in a dialog. 
## PR Checklist

- [x] Tests are passing

## PR Follow ups
- update field editors to make use of this new method
- add documentation for the `openCurrent` method

### DIFF
--- a/lib/dialogs.ts
+++ b/lib/dialogs.ts
@@ -8,6 +8,7 @@ export default function createDialogs(channel, ids) {
     openPrompt: openSimpleDialog.bind(null, 'prompt'),
     openExtension: openExtensionDialog,
     openCurrentApp: openCurrentAppDialog,
+    openCurrent: openCurrentDialog,
     selectSingleEntry: openEntitySelector.bind(null, 'Entry', false),
     selectSingleAsset: openEntitySelector.bind(null, 'Asset', false),
     selectMultipleEntries: openEntitySelector.bind(null, 'Entry', true),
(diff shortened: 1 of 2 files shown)


lib/dialogs.ts
```diff
@@ -23,20 +24,26 @@ export default function createDialogs(channel, ids) {
...
     } else {
       throw new Error('Extension ID not provided.')
     }
   }
 
+  function openCurrentDialog(options?) {
+    if (ids.app) {
+      return openCurrentAppDialog(options)
+    } else {
+      options.id = ids.extension
```
suevalov, line 38: I'd avoid a direct object assignment here cause it changes the value in a reference.

jelz: I think we copy it somewhere in this file before we use ( as we use assignement in the other extension method ).

If not we should fix it.

suevalov:
```
return openExtensionDialog({
...options,
id: ids.extension
})
```

lib/dialogs.ts
```diff
@@ -23,20 +24,26 @@ export default function createDialogs(channel, ids) {
...
+      return openExtensionDialog(options)
+    }
+  }
+
   function openCurrentAppDialog(options?) {
     options = prepareOptions(options)
-
     // Force ID of the current app.
     options.id = ids.app
```
suevalov, line 46: the same

PR #318, opened 2020-05-26
3.15.0

FLoppix:
# Purpose of PR

Release a new version of the sdk

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.14.3",
+  "version": "3.15.0",
```
suevalov, line 4: There should be a change in `package-lock.json` too.

FLoppix: Sorry slipped through :). Good catch!

PR #323, opened 2020-07-16
feat(metadata): Add entry metadata to sdk result

davidfateh:
This adds metadata if it exists to the `sdk.entry` object.

### DIFF
--- a/lib/entry.ts
+++ b/lib/entry.ts
@@ -3,6 +3,7 @@ import { MemoizedSignal } from './signal'
 export default function createEntry(channel, entryData, fieldInfo, createEntryField) {
   let sys = entryData.sys
   const sysChanged = new MemoizedSignal(sys)
+  const entryMetadata = entryData.metadata
 
   channel.addHandler('sysChanged', _sys => {
     sys = _sys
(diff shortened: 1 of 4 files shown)


lib/entry.ts
```diff
@@ -19,6 +20,7 @@ export default function createEntry(channel, entryData, fieldInfo, createEntryFi
     fields: fieldInfo.reduce((acc, info) => {
       acc[info.id] = createEntryField(info)
       return acc
-    }, {})
+    }, {}),
+    ...(entryMetadata ? { metadata: entryMetadata } : {})
```
jelz, line 24: How about always having empty `{}` metadata?

package-lock.json
```diff
@@ -6017,8 +6017,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
```
jelz, line 6020: Can you do `nvm use` before `npm i` and see if we get these?

davidfateh: Fixed.

davidfateh:
> LGTM. Maybe we need a unit test for this?

Added the tests.

PR #330, opened 2020-08-10
feat: add getCurrentState method to uie-sdk app api

shikaan:
## Context

This change is meant to make the App SDK able to fetch the state to be manipulated and provided to the `targetState` portion of the return value of the `onConfig` hook.

### Example flow
1. Get the current state by using `getCurrentState` method
2. Modify the state to introduce/remove/change where the app is to be rendered
3. Return the modified state as `targetState` in the `onConfiguration` callback

[Reference](https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/#register-an-app-configuration-hook)

### To Do
- [x] tests;
- [x] documentation;
- [x] typings;

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -196,6 +196,7 @@ export interface EditorInterface {
     widgetNamespace: string
     settings?: Object
   }
+  editors?: Array<EditorInterface['editor']>
 }
 
 /* Space API */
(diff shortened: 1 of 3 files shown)


jelz:
LGTM but missing all the things around (tests types etc)

lib/types.ts
```diff
@@ -634,11 +635,17 @@ export type PageExtensionSDK = BaseExtensionSDK & {
   ids: Omit<IdsAPI, 'field' | 'entry' | 'contentType'>
 }
 
+export interface AppState {
+  EditorInterface: Record<ContentType['sys']['id'], Omit<EditorInterface, 'sys'>>
```
jelz, line 639: `Omit<EditorInterface, 'sys'>` is not a valid type (e.g. `sidebar: { position: 1}`

PR #331, opened 2020-08-18
Making the EditorWidget non-optional

davidfateh:
This makes the EditorWidget not optional inside the `editors` array.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -177,6 +177,12 @@ export interface ContentType {
   description: string
 }
 
+interface EditorWidget {
+  widgetId: string
+  widgetNamespace: string
+  settings?: Object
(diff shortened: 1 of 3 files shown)


lib/types.ts
```diff
@@ -191,12 +197,8 @@ export interface EditorInterface {
     settings?: Object
     disabled?: boolean
   }>
-  editor?: {
-    widgetId: string
-    widgetNamespace: string
-    settings?: Object
-  }
-  editors?: Array<EditorInterface['editor']>
```
davidfateh, line 199: This code was making `EditorInterface['editor']` optional since it is an optional property.

jelz (approved):

PR #332, opened 2020-08-24
console.error when error happens in app handler.

davidfateh:
# Purpose of PR
The app handlers such as `onConfig` swallow errors and `return false`. I've added `console.error` to keep the functionality the same while exposing errors that happen when these handlers are run for a better debugger experience.

When an error does occur, the handlers `return false` like mentioned about but the installation or save doesn't proceed. This is why I've opted to simply add `console.error` instead of trying to completely change the behavior of these handlers.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/app.ts
+++ b/lib/app.ts
@@ -5,6 +5,11 @@ const isObject = o => typeof o === 'object' && o !== null && !Array.isArray(o)
 const isFunction = f => typeof f === 'function'
 const isPromise = p => isObject(p) && isFunction(p.then)
 
+const handleHandlerError = err => {
+  console.error(err)
+  return false
+}
+
 const runHandler = (handler, defaultResult, handlerArg?) => {
   // Handler was not registered. Registering a handler is not
   // required. We resolve with the default provided in this case.
@@ -18,6 +23,7 @@ const runHandler = (handler, defaultResult, handlerArg?) => {
   try {
     maybeResultPromise = typeof handlerArg === 'undefined' ? handler() : handler(handlerArg)
   } catch (err) {
+    console.error(err)
     return Promise.resolve(false)
   }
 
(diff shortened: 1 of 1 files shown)


kdamball (approved):
As far as the code, looks alright. I'd be interested to know what we missed in the original ticket by opting to go with `onError`

jelz (approved):

lib/app.ts
```diff
@@ -5,6 +5,11 @@ const isObject = o => typeof o === 'object' && o !== null && !Array.isArray(o)
 const isFunction = f => typeof f === 'function'
 const isPromise = p => isObject(p) && isFunction(p.then)
 
+const handleHandlerError = err => {
+  console.error(err)
+  return false
```
shikaan, line 10: What if we replaced

```suggestion
  return Promise.reject(false)
```

and use it also at L26?

shikaan (commented):
Maybe we want to allow customization of error handling function and use `console.error` only if nothing else is defined? Something along the lines of what @jelz suggested in the thread on Slack

lib/app.ts
davidfateh: Because of this: https://github.com/contentful/ui-extensions-sdk/blob/master/lib/app.ts#L63

It expects a handled Promise to come back for it to continue to communicate with the web app. That is why all promises in this function resolve as `false` in the error case.

davidfateh:
> Maybe we want to allow customization of error handling function and use `console.error` only if nothing else is defined? Something along the lines of what @jelz suggested in the thread on Slack

This is something we can put up for a point of discussion. I tested throwing async and sync errors in the callback passed to `init` and found that they do indeed halt the app and show up in the console. To me it makes more sense to assume developers will use `try/catch` when they want to throughout their code.

If we provide our own error handler, we potentially make things more complicated by need to document that there is more than one way to handle errors.

lib/app.ts
shikaan: Yeah, sorry. I meant 

```suggestion
  return Promise.resolve(false)
```

As in L26

davidfateh: What would be the benefit to doing this? As I see it, I'm already resolving false in my error handlers.

shikaan: It's mostly to have one place where we handle this logic, should we decide to have a custom `onError` handler

davidfateh: Oh I see. You're suggesting I replace L26 with this as well and by making the above change I would be able to do that. Got it; can do.

shikaan (approved):

PR #335, opened 2020-08-28
build: use rollup bundler

shikaan:
# Purpose of PR

After getting this https://github.com/contentful/create-contentful-app/issues/15 we have realized the reason for getting that warning is due to how Parcel works internally. 

Given that Parcel [doesn't provide clear guidance about how to deal with the problem](https://github.com/parcel-bundler/parcel/issues/2883), this PR migrates to another build tool.

## Other considered options

* [`esbuild`](https://github.com/evanw/esbuild)
  Configuring esbuild was super easy, incredibly quick and the bundle is eventually smaller. The problem with that is that it is considered an "existence proof", so I wasn't sure to which extent it could be trusted; 

## Tested cases

To make sure this change is not breaking I have replicated locally the cases we expect to have:
* _requiring the package_
  This what bundlers will do and it's been replicated by npm linking the package (this is what `module` in `package.json` is for)
* _using a `<script>`_
  For this I wanted to replicate the behavior of importing the package from `unpkg` (as we do in the ready-made extensions). According to [the documentation](https://unpkg.com/), links to libraries will point to whatever is the target of `main`. Hence, in order to test this, I have run a local server which was serving the main file and it was working.

## Remarks

* I have preferred the `esbuild` experience over this, but it's a risk which I am not sure we want to take;
* We might consider waiting for Parcel or Webpack to fix the issue (and close this), if we don't feel like touching the build;


.nvmrc
```diff
@@ -1 +1 @@
-v13.3.0
\ No newline at end of file
+v13.8.0
```
davidfateh, line 1: Why not use a non-dev version of node?

jelz: Yeah, would be nice use LTS

package.json
```diff
@@ -68,6 +71,8 @@
     "nanoid": "2.1.8",
     "parcel-bundler": "^1.12.4",
```
jelz, line 72: Can we get rid of it?

jelz (approved):

PR #339, opened 2020-09-08
fix: remove entry and contenttype key from baseextensionsdk

Jwhiles:
I noticed that the type for `AppExtensionSDK` doesn't match the underlying implementation, as the type incorrectly includes an `entry` and `contentType` key. This error also applied to some of the other SDK types.

This PR just fixes that type error.

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -8,6 +8,7 @@ import createEditor from './editor'
 import createNavigator from './navigator'
 import createApp from './app'
 import locations from './locations'
+import { BaseExtensionSDK } from './types'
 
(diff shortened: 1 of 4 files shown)


jelz (approved):
@FLoppix once this is published can we update it in your branch and see if it adheres to that? These types are now all correct

FLoppix:
@jelz Yes was thinking the same. I wait once this is published und update everything

FLoppix (approved):

Jwhiles:
@FLoppix @jelz this will actually require changes in felix's branch. I've been working on them though in my branch - so can push that commit to the branch for your review :)

PR #340, opened 2020-09-08
fix: make ids types usable

Jwhiles:
The previous types I introduced were a bit awkward to use. This approach should work. 🤞


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.23.0",
+  "version": "3.24.0",
```
FLoppix, line 4: just a small thing. That should be a  patch version right? So I would say `3.23.1` is more appropriate?

FLoppix (approved):

PR #342, opened 2020-09-10
Prefer Contentful Apps to UI Extensions

kdamball:
# Purpose of PR

Trying to replace, wherever possible, UI Extensions with Apps


jelz:
was it consulted with @fabe ? for this library, which is still the only option for UIE development I'd still mention both UIEs and Apps

kdamball:
Nah, haven't spoken to him - I assumed it's part of the whole extension to apps mvmt. I had him as part of reviewers.

kdamball:
Need more direction - closed it for now

PR #345, opened 2020-10-13
Fix/finish typescript migration

VaguelySerious:
# Finish typescript migration

## Changed the following
- compile w/ strict = true
- implement interfaces where possible

## These changes seem to have already been in effect
- make private methods private
- ES6 modules
- run tests in ts-node
- allow explicit any in PUBLIC interface

## PR Checklist

- [ x ] Tests are added/updated/not required
- [    ] Tests are passing (unit tests yes, still waiting on integration running on CircleCI)
- [ x ] Typescript typings are added/updated/not required


VaguelySerious:
I accidentally included @kdamball 's commit that updates the naming in the README. Please let me know if I should try to pull that out again or if it's fine.

Jwhiles (approved):
This looks like a good improvement to me! I still see some room for making these types more specific - I think a lot of the `Function` arguments could be made more specific for example. But maybe that can wait for a future PR :)

lib/types.ts
```diff
@@ -595,7 +639,7 @@ export interface BaseExtensionSDK {
   parameters: ParametersAPI
   /** Exposes method to identify extension's location */
   location: LocationAPI
-  /** Exposes methods for checking user's access level */
+  /** Exposes metmessagehods for checking user's access level */
```
jelz, line 642: typo

package.json
```diff
@@ -13,8 +13,8 @@
   ],
   "scripts": {
     "test": "ts-mocha 'test/unit/*.[jt]s'",
-    "lint": "tsc -b && eslint '{lib,test}/**/*.{t,j}s'",
-    "lint:fix": "eslint --fix lib/ test/",
+    "lint": "eslint '{lib,test}/**/*.{t,j}s'",
+    "lint:fix": "npm run lint -- --fix",
     "build": "npm run typings && rollup -c --compact",
     "build:debug": "npm run build -- --sourcemap",
     "typings": "tsc -b typings.tsconfig.json && mv lib/types.d.ts typings.d.ts && cat module-declaration.d.ts >> typings.d.ts",
```
jelz, line 20: Could we eliminate typings somehow? Doesn't make sense to have them in 100% TS project.

lib/app.ts
```diff
@@ -1,16 +1,19 @@
+import { Channel } from './channel'
+import { AppConfigAPI, AppState } from './types'
+
 const HOOK_STAGE_PRE_INSTALL = 'preInstall'
 const HOOK_STAGE_POST_INSTALL = 'postInstall'
 
-const isObject = o => typeof o === 'object' && o !== null && !Array.isArray(o)
-const isFunction = f => typeof f === 'function'
-const isPromise = p => isObject(p) && isFunction(p.then)
+const isObject = (o: any) => typeof o === 'object' && o !== null && !Array.isArray(o)
```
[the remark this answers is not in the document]
VaguelySerious: I wouldn't move it to a future PR, because the ticket likely wouldn't be worth the overhead of checking all the types again. I'll try using type guards.
On that note: I think there's a few tests that check for correct types on return etc. that aren't doing much since we're using typescript. Should I remove those?

package.json
VaguelySerious: I did some testing on this. We'd still have a separate command to generate typings, but using the source files instead of the `types.ts` file. The types would still mostly depend on the `types.ts` file as it is heavily imported throughout the code.
I'm currently making sure that the developer experience, especially autocomplete features, work exactly the same.

VaguelySerious: I think splitting it into two PRs makes sense for this issue. The changes for this can be made irrespective of the rest of the work done in this PR without any overhead, and I imagine it is better to get the changes out quicker so other people can work in the complete typescript context @jelz What do you think?

VaguelySerious:
@djagya We identified a problem in `user_interface` that causes integration tests on this PR to fail. We decided to fix the behavior in `user_interface` before merging this. See https://github.com/contentful/user_interface/pull/6968

VaguelySerious:
@djagya Just a heads up: We identified a problem in the web app that causes integration tests on this PR to fail. We decided to fix the behavior in web app first before merging this.

Issue #321, opened 2020-06-11
navigator.openEntry resolve promise on close


joeweale:
Hi,

Having trouble with waitForClose on the navigator sdk. It doesnt seem to resolve if the item is deleted but works fine if it's not.

I'd expect it to still resolve but with an empty entry property.

Thanks

Joe

Jwhiles:
Hey Joe, thanks for reporting this issue. I've been able to reproduce the problem, and will record the bug.

VaguelySerious:
Hi Joe, thank you again for bringing this to our attention. `sdk.navigator.openEntry` with `waitForClose` now resolves if the entity is deleted by any user during use of the slide-in editor.
Please refer to the [updated documentation](https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/#navigator) as to the exact details.

joeweale:
Thanks all 👍

PR #348, opened 2020-10-21
fix: make entity optional on navigator response

VaguelySerious:
# Purpose of PR

As mentioned in #321, if an entry is deleted during a `navigator.openEntry` call with the `waitForClose` property, the promise is not resolved.
We should not resolve with an outdated entry, but also can not resolve with an nonexistent entry. For this reason we need to change the promise return type to allow for a missing entry in case it has been deleted during user interaction.

## PR Checklist

- [x] No update to tests are required, typing change only
- [x] Tests are passing
- [x] Typescript typings are updated

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -509,7 +509,7 @@ export interface NavigatorSlideInfo {
 
 export interface NavigatorOpenResponse<T> {
   navigated: boolean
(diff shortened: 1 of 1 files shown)


Jwhiles (approved):
:+1:

PR #390, opened 2020-10-29
feat: log error if it appears app is running outside of iframe

Jwhiles:
This PR tries to make it more obvious what's going wrong when ui-extension-sdk is used outside of Contentful. It shows an error when init is called outside of an iframe

image

The link doesn't currently point to a specific question in the FAQ - I intend to update the FAQ with that information.

### DIFF
--- a/lib/initialize.ts
+++ b/lib/initialize.ts
@@ -22,8 +22,21 @@ export default function createInitializer(
 
   return function init(
     initCb: (sdk: KnownSDK, customSdk: any) => any,
-    { makeCustomApi }: { makeCustomApi?: Function } = {}
+    {
+      makeCustomApi,
+      supressIframeWarning
+    }: { makeCustomApi?: Function; supressIframeWarning?: boolean } = {
+      supressIframeWarning: false
+    }
   ) {
+    if (!supressIframeWarning && currentWindow.self === currentWindow.top) {
+      console.error(`Cannot use ui-extension-sdk outside of Contenful:
+
+In order for the ui-extension-sdk to function correctly, your app needs to be run in an iframe in the Contentful Web App.
+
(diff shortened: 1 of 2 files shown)


lib/initialize.ts
```diff
@@ -22,8 +22,23 @@ export default function createInitializer(
...
+    }: { makeCustomApi?: Function; supressIframeWarning?: boolean } = {
+      supressIframeWarning: false
+    }
   ) {
+    if (!supressIframeWarning && currentWindow.self === currentWindow.top) {
+      console.error(`Cannot use ui-extension-sdk outside of Contenful:
+
+It appears that you are trying to use the ui-extension-sdk outside of Contentful
+
+In order for the ui-extension-sdk to function correctly, your app needs to be run in an iframe in the Contentful webapp
```
andipaetzold, line 37: In the FAQ we use "Web App" instead of webapp.

```suggestion
In order for the ui-extension-sdk to function correctly, your app needs to be run in an iframe in the Contentful Web App
```

lib/initialize.ts
```diff
@@ -22,8 +22,23 @@ export default function createInitializer(
...
+
+Learn more about local development with the ui-extension-sdk here:
+  https://www.contentful.com/developers/docs/extensibility/ui-extensions/faq/#local-development-using-the-sdk`)
+    }
```
andipaetzold, line 41: I am not 100% sure about the wording. Content wise it's fine. Maybe add some periods after the sentences.

andipaetzold (approved):

VaguelySerious (approved):
I'm with Andi on the naming ("web app") and the punctuation. I don't think the newlines in the log message are necessary either, but that's a style choice.

PR #392, opened 2020-11-05
feat: add request signing endpoint in space API

VaguelySerious:
# Purpose of PR

Expose `sdk.space.signRequest` and add type to SpaceAPI

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/space.ts
+++ b/lib/space.ts
@@ -1,5 +1,6 @@
 import { ContentType, SpaceAPI } from './types'
 import { Channel } from './channel'
+import { IdsAPI } from '../typings'
 
 const spaceMethods: Array<keyof SpaceAPI> = [
   'getContentType',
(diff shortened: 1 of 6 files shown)


lib/types.ts
```diff
@@ -316,6 +316,23 @@ export interface SpaceAPI {
   ) => Promise<ScheduledAction[]>
   /* Returns a list of scheduled actions for the currenst space & environment */
   getAllScheduledActions: () => Promise<ScheduledAction[]>
+
+  signRequest?: (request: CanonicalRequest) => Promise<AppSignedRequest>
```
VaguelySerious, line 320: Alternatively I could make it return just the `additionalHeaders` instead of wrapping it together with the unneeded `sys`. This would also save us from redefining the type for the `AppSignedRequest` entity here.

jelz: I would do so, especially given the fact that there's more in `sys` compared to what you've defined below.

lib/space.ts
```diff
@@ -44,7 +44,9 @@ const spaceMethods: Array<keyof SpaceAPI> = [
   'getUsers',
 
   'getAllScheduledActions',
-  'getEntityScheduledActions'
+  'getEntityScheduledActions',
+
+  'signRequest'
```
jelz, line 49: Don't we want to make it available for apps only?

jelz (approved):

PR #393, opened 2020-11-06
Bump dependencies

andipaetzold:
Manually bump dependencies to prepare for automatic dependency updates with dependabot.

This only bumps the dependencies of the main `package.json`. Not of the test-extension's.

After merge https://github.com/contentful/ui-extensions-sdk/pull/389 can be merged as well to enable dependabot.

What actually changed?

- Reformatting because of updated prettier
- New TypeScript requires `new Promise<void>` for promises without resolved value
- Reordering typings as we enforce that a type can only be used AFTER it has been defined
- Cypress update to 5.0 with `blacklistHosts` -> `blockHosts`
- Remove `sourceMap` and `inlineSourceMap` from `tsconfig.json` as we don't want to emit them to npm and also already manually enable them with `build:debug`. This caused an issue when cypress ran its webpack build.


lib/types.ts
```diff
@@ -284,7 +304,7 @@ export interface SpaceAPI {
   ) => Promise<CollectionResponse<T>>
 
   getAsset: <T = Object>(id: string) => Promise<T>
-  getAssets: <T = Object, InputArgs = SearchQuery>(
+  getAssets: <T = Object, InputArgs = SearchQuery>( // eslint-disable-line @typescript-eslint/no-unused-vars
```
Jwhiles, line 307: We should change this to match getEntries on line 292 like so 
```suggestion
  getAssets: <T = Object, InputArgs = SearchQuery>(query?: InputArgs) => Promise<CollectionResponse<T>>
```

or potentially just remove `InputArgs` entirely, and 
```suggestion
  getAssets: <T = Object>(query?: SearchQuery) => Promise<CollectionResponse<T>>
```

Jwhiles (commented):
Looks good! I spotted one opportunity to remove a lint-disable line, but other than that I think this is good to merge :)

lib/types.ts
andipaetzold: Yes, I agree that we should remove the `eslint-disable-line`. I added it because eslint showed a warning for `InputArgs` being unused which doesn't make sense in this case as it's a type definition.

andipaetzold: Screenshot 2020-11-23 at 15 34 52

Jwhiles (approved):

PR #431, opened 2020-12-09
feat: [EXT-2360] add test for onvaluechanged

Jwhiles:
# Purpose of PR

## PR Checklist

- [x] Tests are passing


test/cypress/integration/reusable/entry-value-changed-callback-test.ts
```diff
@@ -0,0 +1,18 @@
+export function entryValueChangedCallbackTest(iframeSelector: any) {
+  it('verifies on valueChanged is called when value changes', () => {
+    cy.getSdk(iframeSelector).then(async (sdk) => {
+      const spy = cy.stub()
+      sdk.entry.fields.body.onValueChanged(spy)
+      expect(spy).to.be.calledOnce
+
+      cy.get('@extension').within(() => {
```
Jwhiles, line 8: I'm not sure if it makes sense to actually change the entry this way, really we're just testing the extension at that point. Arguably it's enough to test that `onValueChanged` was called on initial render. WDYT?

FLoppix: You when the extension was rendered?

Jwhiles: Yeah, exactly

FLoppix: Then I agree with you that we should test only on render. We should not alter the state of the extension or interact with it.

kdamball: For clarification, was the original issue that it wasn't even called on initial render?

FLoppix (approved):

FLoppix (approved):

PR #455, opened 2020-12-22
chore: enable semantic release

andipaetzold:
Currently, the package is "only" deployed as `contentful-ui-extensions-sdk`

Deploy package as three different packages:
* `contentful-ui-extensions-sdk`
* `@contentful/ui-extensions-sdk`
* `contentful-app-sdk`
* `@contentful/app-sdk`

The script is copied and adjusted from https://github.com/contentful/create-contentful-app/blob/master/scripts/deploy.js

Also, the publishing now happens automatically using `semantic-release`


scripts/shared.js
```diff
@@ -0,0 +1,37 @@
+const path = require('path')
+const spawn = require('cross-spawn')
+const fs = require('fs')
+
+const PACKAGES = [
+  'contentful-ui-extensions-sdk',
+  '@contentful/ui-extensions-sdk',
+  '@contentful/app-sdk',
```
kdamball, line 8: We can also add `contentful-app-sdk`

package.json
```diff
@@ -26,12 +26,16 @@
     "cypress:run": "cypress run --browser chrome",
     "clean-reports": "rm -rf cypress/reports",
     "merge-report": "mochawesome-merge ./test/cypress/reports/separate-reports/*.json > ./test/cypress/reports/full_report.json",
-    "generate-report": "npx mochawesome-report-generator --reportDir test/cypress/reports test/cypress/reports/full_report.json"
+    "generate-report": "npx mochawesome-report-generator --reportDir test/cypress/reports test/cypress/reports/full_report.json",
+    "semantic-release": "semantic-release"
```
kdamball, line 30: Shouldn't we add a publish script also? e.g: `npm run publish-all` similar to cca

scripts/shared.js
andipaetzold: I assumed that the library was first published as `contentful-ui-extensions-sdk` and then `@contentful/ui-extensions-sdk` was added when npm scopes were introduced. I therefore now only "only" added the scoped version of `app-sdk`.

package.json
andipaetzold: I Will add one, even though we hopefully won't use it after the first publish

scripts/shared.js
kdamball: It's more about package name parking to avoid copy cats/name squatting.

andipaetzold: Makes sense. Just added it.

kdamball (approved):

PR #490, opened 2021-01-13
feat: export module as contentfulApp

andipaetzold:
If you don't use a bundler you can simply import the library in your html file and use it with 

```javascript
window.contentfulExtension.init(...)
```

We want to make it possible to _also_ use `window.contentfulApp`. Unfortunately, it is not possible to export multiple variables using rollup. A workaround is using the `footer` property:
> Code to insert at end of bundle (outside wrapper)

This seems to be the recommended approach: https://github.com/rollup/rollup/issues/553#issuecomment-202166713

I tested it with the following code:
```html
    <script src="dist/cf-extension-api.js"></script>
    <script>
        console.log(window.contentfulApp);
        console.log(window.contentfulExtension);
    </script>
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -21,6 +21,7 @@ module.exports = [
   makeConfigForOutput({
     file: pkg.main,
     format: 'umd',
-    name: 'contentfulExtension'
+    name: 'contentfulExtension',
+    footer: 'window.contentfulApp = window.contentfulExtension;'
   })
 ]


rollup.config.js
```diff
@@ -21,6 +21,7 @@ module.exports = [
   makeConfigForOutput({
     file: pkg.main,
     format: 'umd',
-    name: 'contentfulExtension'
+    name: 'contentfulExtension',
+    footer: 'window.contentfulApp = window.contentfulExtension;'
```
jelz, line 25: Nice, `footer`. How about `;window.contentfulApp....`?

jelz (commented):
Can you also test with `window.contentfulApp === window.contentfulExtension`? ;)

andipaetzold:
> Can you also test with `window.contentfulApp === window.contentfulExtension`? ;)

-> `true`

rollup.config.js
andipaetzold: As we run terser over the whole thing, the minified JavaScript is identical.

The footer is inserted after the wrapping function which already has a semicolon at its end.

```javascript
[...]
    };
}var init = createInitializer(window, createAPI);exports.init=init;exports.locations=locations;Object.defineProperty(exports,'__esModule',{value:true});})));
window.contentfulApp = window.contentfulExtension;
```

jelz (approved):

PR #516, opened 2021-01-28
feat: add Task API to sdk.entry

VaguelySerious:
# Purpose of PR

Adds methods that allow accessing/updating `Task` entities on an `Entry`

## PR Checklist

- [x] Tests are passing


package.json
```diff
@@ -136,5 +136,6 @@
       ],
       "@semantic-release/github"
     ]
-  }
+  },
+  "dependencies": {}
```
andipaetzold, line 140: Just curious: Is this change required?

lib/types.ts
```diff
@@ -131,7 +139,44 @@ export interface EntryFieldAPI {
...
+  sys: TaskSys
+}
+
+/** Allows accessing the Task API for the current entry. */
+export interface TaskAPI {
+  getTask(id: string): Promise<Task>
+  getTasks(): Promise<CollectionResponse<Task>>
+  createTask(data: TaskInputData): Promise<Task>
+  updateTask(task: Task): Promise<Task>
+  deleteTask(task: Task): Promise<void>
```
andipaetzold, line 176: The delete calls in `SpaceAPI` seem to return the actual object that was deleted.
Any reason for not doing that here?

lib/types.ts
```diff
@@ -131,7 +139,44 @@ export interface EntryFieldAPI {
...
+}
+
+export interface EntryAPI extends TaskAPI {
```
andipaetzold, line 179: Not happy with extending `EntryAPI` here as we can only extend from one interface. This will become problematic when adding new apis to the entry in the future.

I'd rather add all functions of `TaskAPI` directly into `EntryAPI`

test/unit/entry.spec.ts
```diff
@@ -65,6 +70,24 @@ describe('createEntry()', () => {
...
+          const args = ['foo', 42, {}]
+          describeChannelCallingMethod({
+            creator: createEntry,
+            methodName,
+            channelMethod: 'callEntryMethod',
+            args: args,
+            expectedCallArgs: [methodName, args],
+            makeCreatorArgs: () => {
+              return [{}, [{}], () => ({})]
+            },
```
andipaetzold, line 85: `makeCreatorArgs` is confusing to me.
How about

```javascript
creator: (channel) => createEntry(channel, {}, [{}], () => ({}))
```

and removing that again. It feels like it makes it unnecessarily complex

andipaetzold (changes requested):
This should work. Still some comments.

Also, we need to add some integration tests before mergin.

lib/types.ts
VaguelySerious: `interface A extends B, C, D {}` works?
And the reason for the split is that it made it easier to add the methods in `entry.ts`
```js
  const taskApi = {} as TaskAPI
  ...
  return {
    otherThings,
   ...taskApi
   }
```
But it's also easily possible to do this without splitting the interface, so if you still feel strongly about changing it let me know

andipaetzold: Oh - didn't know extending multiple interfaces

package.json
VaguelySerious: No, but npm added this for me, so I assume it's good practice

lib/types.ts
VaguelySerious: I was going off of the CMA docs. This seems to be a mistake either in the docs, or in the SpaceAPI then, good catch! I'll investigate

VaguelySerious: So this code is ok for now, but we'll need to change SpaceAPI later

test/unit/space.spec.ts
```diff
@@ -57,6 +57,7 @@ describe('createSpace()', () => {
         channelMethod: 'callSpaceMethod',
         args: args,
         expectedCallArgs: [spaceMethod, args],
+        makeCreatorArgs: () => [],
```
Jwhiles, line 60: Can this now be removed? Looks like you removed the associated code.

test/cypress/integration/sidebar-extension.spec.ts
```diff
@@ -71,5 +77,30 @@ context('Sidebar extension', () => {
     })
   })
 
+  it('creates, gets, and deletes a content type through spaceAPI', () => {
```
VaguelySerious, line 80: I didn't want to add _all_ the methods for _all_ the entities on space API, because we're not testing the CMA implementation, instead tried to get a full basic example in, so we at least know the integration is working on an SDK level.

Jwhiles (approved):

test/cypress/integration/sidebar-extension.spec.ts
```diff
@@ -71,5 +77,27 @@ context('Sidebar extension', () => {
...
+    cy.getSdk(iframeSelector).then(async (sdk: SidebarExtensionSDK) => {
```
andipaetzold, line 81: I personally don't like the mix of `.then` and `async`. Either one or the other.

andipaetzold (approved):

test/cypress/integration/sidebar-extension.spec.ts
VaguelySerious: It's not possible to `await` the `getSdk` call, sorry!

PR #639, opened 2021-04-06
feat: expose tags api

damienxy:
# Purpose of PR

Expose the tags API so it can be used for frontend apps

## PR Checklist

- [x] Tests are passing


Jwhiles (approved):
Looks good to me!

lib/space.ts
```diff
@@ -47,6 +47,11 @@ const spaceMethods: Array<keyof SpaceAPI> = [
   'getEntityScheduledActions',
 
   'signRequest',
+
+  'createTag',
+  'readTags',
```
hwartig, line 52: @damienxy I'm surprised to see `readTags` instead of `getTags` here given that we have `get{Users|Entries|...}` What am I missing?

damienxy: It is because the implementation uses the `tagsRepo` from the user interface, not directly the CMA.

PR #681, opened 2021-05-05
fix: make options of dialogs.openExtension optional

andipaetzold:
Currently, it is required to pass an object when calling `openExtension`. As it is valid to pass an empty object `{}`, I made options optional. That way the signature matches `openCurrentApp` and `openCurrent`.

This is a `fix` as it probably should always have been that way. Also `feat` feels to big for this change.


Jwhiles (approved):
Seems legit

PR #699, opened 2021-05-10
fix: include typings in the release

shikaan:


lib/types/space.types.ts
```diff
@@ -94,7 +93,11 @@ export interface SpaceAPI {
   signRequest: (request: CanonicalRequest) => Promise<Record<string, string>>
 
   createTag: (id: string, name: string, visibility?: TagVisibility) => Promise<Tag>
+  /**
+   * @deprecated use `getTags` instead
+   */
   readTags: (skip: number, limit: number) => Promise<CollectionResponse<Tag>>
+  getTags: (skip: number, limit: number) => Promise<CollectionResponse<Tag>>
```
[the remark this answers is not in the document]
shikaan: When alpha will be stable they'll work, might as well add them now. WDYT?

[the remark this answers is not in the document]
andipaetzold: I'd only add the new types when doing the actual ticket. We currently don't have a timeline for the renaming.

lib/types/utils.ts
```diff
@@ -2,11 +2,11 @@ export type WithOptionalSys<Type extends { sys: unknown }> = Omit<Type, 'sys'> &
   sys?: Type['sys']
 }
 
-export interface Link<Type = string> {
+export interface Link<LinkType = string, Type = string> {
```
andipaetzold, line 5: Can we make `Link` the default?

```suggestion
export interface Link<LinkType = string, Type extends string = 'Link`> {
```

And why not making `Type` a fixed value as in cma.js 
https://github.com/contentful/contentful-management.js/blob/ebc7af28d184db031fad8179b24cab2d4fbc5bb6/lib/common-types.ts#L70-L76

lib/types/utils.ts
```diff
@@ -18,7 +18,7 @@ export interface CollectionResponse<T> {
   sys: { type: string }
 }
 
-export type ContentEntityType = 'Entry' | 'Asset'
+export type ContentEntityType = 'Entry' | 'Asset' | string
```
andipaetzold, line 21: Why?

lib/types/utils.ts
andipaetzold: Oh... great - let's keep it like this

PR #710, opened 2021-05-12
feat: test that SDK respects alias

Jwhiles:
Adds a test for the entry editor location to make sure the SDK is hitting the correct endpoints in aliased environments.

If we're happy with this approach, we might want to add similar tests for each SDK/location as a follow up

### DIFF
--- a/test/integration/index.ts
+++ b/test/integration/index.ts
@@ -114,6 +114,7 @@ const run = async () => {
     environmentId: tempEnvironmentId,
     role: 'admin',
     entries: entryIds,
+    aliasId: tempAliasId,
   })
   await runCypress('admin')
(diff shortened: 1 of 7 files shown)


test/integration/index.ts
```diff
@@ -127,6 +141,7 @@ const run = async () => {
     environmentId: 'master-test',
     role: 'editorMasterOnly',
     entries: newEntryIds,
+    aliasId: '',
```
Jwhiles, line 144: not sure about this, gonna dig into what this config is used for

andipaetzold (approved):
Looks good. Seems like a straightforward way to check forward way to verify request paths.

PR #711, opened 2021-05-12
feat/more explicit space api types

shikaan:


lib/types/utils.ts
```diff
@@ -1,5 +1,13 @@
-export type WithOptionalSys<Type extends { sys: unknown }> = Omit<Type, 'sys'> & {
-  sys?: Type['sys']
+type SysWithId = {
+  id: string
+}
+
+export type WithOptionalId<Type extends { sys: unknown }> = Omit<Type, 'sys'> & {
+  sys?: Omit<Type['sys'], 'id'> & Partial<SysWithId>
```
andipaetzold, line 6: I am not sure how perfect you want the types to be, but shouldn't we remove `Omit<Type['sys'], 'id'> &`?

Now, if we want to set an id for a new entry, we would also need to specify `version`, `updatedAt` etc. Also the `contentType` link should be removed as that's set with a different argument of the `createEntry` call.

```suggestion
  sys?: Partial<SysWithId>
```

shikaan: Keeping the first part allows the generic `Type` to specify the rest of the `sys` without the need of allowing everything with index access. It looked like a good balance between flexibility and strictness.

The case for setting an id upon creation is with ContentType (and in an ideal world where this whole interface can be made consistent also with Tags). In other cases we do not expect the id to be changed nor set: as a matter of fact we ignore the `sys.id` and this is what I was trying to capture here.

So - given that the other properties are not the same for each and every entity - we allow the generic `Type` to be set, but when we want an optional Id (for example in cases when - depending on the presence of an id the action becomes a PUT or a POST) we just enforce that.

Does it answer the question? 🤔

shikaan: @andipaetzold  <- just to get you a notification :P

kdamball (approved):

PR #722, opened 2021-05-18
feat: expose WithId

shikaan:
# Purpose of PR

* expose `withId`
* capture the following cases with typings:
  * create with only `sys.id` specified (e.g. Content Type)
  * create with either at least `sys.id` or no `sys` at all (e.g. Entry)

### DIFF
--- a/lib/types/utils.ts
+++ b/lib/types/utils.ts
@@ -1,14 +1,19 @@
-type SysWithId = {
+type Sys = {
   id: string
+  type: string
+  [key: string]: any
 }
 
-export type WithOptionalId<Type extends { sys: unknown }> = Omit<Type, 'sys'> & {
-  sys?: Partial<Omit<Type['sys'], 'id'> & SysWithId>
+type Entity = {
+  sys: Sys
+  [key: string]: any
 }
 
-export type WithId<Type extends { sys: unknown }> = Omit<Type, 'sys'> & {
-  sys: Partial<Omit<Type['sys'], 'id'>> & SysWithId
-}
+type Optional<Type, Keys extends keyof Type> = Partial<Type> & Omit<Type, Keys>
+
+export type WithSysWithAtLeastId = { sys: { id: string; [key: string]: any } }
+export type WithId<Type extends Entity> = Omit<Type, 'sys'> & WithSysWithAtLeastId
+export type WithOptionalId<Type extends Entity> = Optional<Type, 'sys'> | WithId<Type>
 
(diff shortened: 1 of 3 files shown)


lib/types/utils.ts
```diff
@@ -1,14 +1,19 @@
-type SysWithId = {
+type Sys = {
   id: string
+  type: string
+  [key: string]: any
```
kdamball, line 4: Equivalent to `{id: string; type: string;} & Record<string, any>` ?

lib/types/utils.ts
```diff
@@ -1,14 +1,19 @@
...
+  sys: Sys
+  [key: string]: any
 }
 
-export type WithId<Type extends { sys: unknown }> = Omit<Type, 'sys'> & {
-  sys: Partial<Omit<Type['sys'], 'id'>> & SysWithId
-}
+type Optional<Type, Keys extends keyof Type> = Partial<Type> & Omit<Type, Keys>
+
+export type WithSysWithAtLeastId = { sys: { id: string; [key: string]: any } }
```
kdamball, line 14: Can you have `sys` without `type`?

lib/types/utils.ts
andipaetzold, line 4: This is not needed. The same applies to line 9.

lib/types/utils.ts
```diff
@@ -1,14 +1,19 @@
...
+export type WithId<Type extends Entity> = Omit<Type, 'sys'> & WithSysWithAtLeastId
```
andipaetzold, line 15: This drops all properties of `Type.sys` except of `id`. I'd therefore do something like this and remove line 14.

```suggestion
export type WithId<Type extends Entity> = Omit<Type, 'sys'> & { sys: Omit<Entity['sys'], 'id'> & Required<Pick<Entity['sys'], 'id'>> }
```

That way the properties don't get lost.

andipaetzold (changes requested):

lib/types/utils.ts
shikaan: Yes. Or rather, for ContentType creation (the case this was created for) you can specify a type, but it's really useless information we ignore, hence the indexed access

lib/types/utils.ts
shikaan: Well, in this case we purposely want the properties to get lost because we ignore them in the underlying calls (this also answers the the comment above about using string-indexed access)

andipaetzold (approved):

PR #752, opened 2021-05-28
test: adds integration alias tests

loweisz:
Bringing back the alias tests from https://github.com/contentful/ui-extensions-sdk/pull/710

The difference is that we don't create a new alias on every test run as this is not really needed for those tests. 
This means we are heavily relying on an existing and predefined alias within the test space. 

To make sure that it won't fail it also includes a preflight check for the existence of this alias and creates it if not available.


test/integration/tasks/create-new-environment.ts
```diff
@@ -33,5 +33,26 @@ export default async () => {
...
+    if (JSON.parse(e.message).status === 404) {
+      console.log(`"${testAliasId}" is not available, lets try to create it`)
+      try {
+        const baseEnvironment = await space.getEnvironment('test-base')
+        await space.createEnvironmentAliasWithId(testAliasId, {
+          environment: baseEnvironment as any,
+        })
+        console.log(`New alias "${testAliasId}" has been created`)
+      } catch (e) {
+        console.log('Could not create the alias', e)
```
Jwhiles, line 50: should we rethrow here? Seems like the test run will fail if the Alis is missing

loweisz: Yes that actually makes a lot of sense 👍

andipaetzold (approved):

PR #765, opened 2021-06-04
fix: update onConfigure type to match reference

andipaetzold:
> The callback provided can be synchronous or asynchronous.

> If the callback returns false, throws or returns a promise which will be eventually rejected, the process is aborted and no changes are made. This way the app can validate its state and user inputs before continuing.

### DIFF
--- a/lib/types/app.types.ts
+++ b/lib/types/app.types.ts
@@ -25,10 +25,10 @@ export interface AppConfigAPI {
   getParameters: <T extends KeyValueMap = KeyValueMap>() => Promise<null | T>
   /** Registers a handler to be called to produce parameters for an App */
   onConfigure: (
-    handler: () => {
-      parameters?: KeyValueMap
-      targetState?: AppState
-    }
+    handler: () =>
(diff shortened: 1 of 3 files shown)


lib/types/app.types.ts
```diff
@@ -25,10 +25,10 @@ export interface AppConfigAPI {
   getParameters: <T extends KeyValueMap = KeyValueMap>() => Promise<null | T>
   /** Registers a handler to be called to produce parameters for an App */
   onConfigure: (
-    handler: () => {
-      parameters?: KeyValueMap
-      targetState?: AppState
-    }
+    handler: () =>
+      | Promise<{ parameters?: KeyValueMap; targetState?: AppState }>
+      | { parameters?: KeyValueMap; targetState?: AppState }
```
kdamball, line 30: makes sense to make it it's own type and reuse it here and the other file. Considering it's even explained in docs

andipaetzold: makes sense. Also going to export it to simplify typing the actual app

kdamball (approved):
🎉

PR #789, opened 2021-06-14
feat: add interface for space.getTeams [HOMER-75]

Chaoste:
# Purpose of PR

In order to expose teams for the tasks app, the AppSDK exposes all available teams within a space.

[Next PR](https://github.com/contentful/user_interface/pull/8376)

## PR Checklist

- [x] Tests are passing


lib/types.ts
```diff
@@ -36,6 +36,32 @@ export interface Link {
...
+    archivedVersion?: number
+    archivedBy?: Link
+    archivedAt?: string
+    deletedVersion?: number
+    deletedBy?: Link
+    deletedAt?: string
+    id: string
+    type: string
+    memberCount: number
+    organization: {
```
ronaldronson, line 53: Should it also be a `Link` type?

lib/types.ts
```diff
@@ -36,6 +36,32 @@ export interface Link {
   }
 }
 
+export interface Team {
+  sys: {
+    space?: Link
+    status?: Link
```
ronaldronson, line 42: Is the `status` really an entity?

giotiskl: status is not an entity and it's not in `sys` it's a top-level field outside of the `sys` object

Chaoste: Sounds reasonable. I just copied the type from cma ([Source](https://github.com/contentful/contentful-management.js/blob/master/lib/entities/team.ts#L7)):
```
export type TeamProps = {
  sys: MetaSysProps & { memberCount: number; organization: { sys: MetaLinkProps } }
  name: string
  description: string
}
```
The [MetaSysProps defines the status as a link](https://github.com/contentful/contentful-management.js/blob/master/lib/common-types.ts#L141). I updated this type but do one of you know why the cma is typing status as a link?

ronaldronson (approved):
LGTM

PR #807, opened 2021-06-23
feat: add type parameter to create task input data

ronaldronson:
# Purpose of PR

Add the `assignedToType` filed to the TaskInputData so the task could be assigned to other entity types but the User.
Make this field optional for backward compatibility.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/lib/types/entry.types.ts
+++ b/lib/types/entry.types.ts
@@ -6,7 +6,7 @@ type TaskState = 'active' | 'resolved'
 
 export interface TaskInputData {
   assignedToId: string
-  assignedToType?: string
+  assignedToType?: 'User' | 'Team'
   body: string
   status: TaskState
   dueDate?: string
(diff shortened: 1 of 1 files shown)


lib/types/entry.types.ts
```diff
@@ -6,6 +6,7 @@ type TaskState = 'active' | 'resolved'
 
 export interface TaskInputData {
   assignedToId: string
+  assignedToType?: string
```
FLoppix, line 9: can we limit the type to only `User` or `Team` right now? Also better DX for auto completion

ronaldronson: Good point.

FLoppix (approved):
LGTM. I assume we always make sure that we fallback to a `User` when we do not specifiy a `assignedToType` ?

PR #809, opened 2021-06-23
feat: add type parameter to create task input data

ronaldronson:
# Purpose of PR

Add the assignedToType filed to the TaskInputData so the task could be assigned to other entity types but the User.
Make this field optional for backward compatibility.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -191,6 +191,7 @@ export interface TaskSys {
 
 export interface TaskInputData {
   assignedToId: string
+  assignedToType?: 'User' | 'Team'
   body: string
   status: TaskState
   dueDate?: string


lib/types.ts
```diff
@@ -191,6 +191,7 @@ export interface TaskSys {
 
 export interface TaskInputData {
   assignedToId: string
+  assignedToType?: 'User' | 'Team'
```
giotiskl, line 194: why does this take an `assignedToId` that is a `string` as a value here? :/ I'd expect this to be consistent with the public interface of the API and have instead an `assignedTo: Link<User | Team>`

giotiskl: i guess it'd be a breaking change so this is probably the best we can do

giotiskl (approved):

lib/types.ts
giotiskl: @FLoppix do we want to do lowercase or uppercase for `assignedToType`?

PR #816, opened 2021-06-28
feat: [] add cma adapter to sdk

loweisz:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing


lib/api.ts
```diff
@@ -47,6 +48,7 @@ function makeSharedAPI(channel: Channel, data: ConnectMessage): BaseExtensionSDK
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
   return {
+    cmaAdapter: createAdapter(channel),
```
Jwhiles, line 51: We were originally planning to just call this `adapter` - right?

lib/types.ts
```diff
@@ -705,6 +707,8 @@ export interface BaseExtensionSDK {
   access: AccessAPI
   /** Exposes relevant ids, keys may be ommited based on location */
   ids: IdsAPI
+  /** the adapter that can be used for creating the cma client */
```
Jwhiles, line 710: Not sure this comment is correct, the adapter is already created at the time we create the SDK, right?

loweisz: Yes, the adapter is created, but the comment says that you can use it to create the cma client, or did I word that wrong?🤔.

kdamball (approved):

lib/api.ts
andipaetzold: Yes, we thought the name `adapter` is a little generic and therefore went for `cmaAdapter` in case there will be more adapters in the future (e.g. GraphQL adapter?).
Also, the name anyway doesn't match with the `cma.js` as it needs to passed in there as `apiAdapter`

lib/types.ts
andipaetzold: The comment is correct but still should be reworded to be consistent with the other comments

shikaan (approved):
LGTM. 

I would only (super nit) rephrase the comment as 

"Adapter to be injected in `contentful-management` client" 

this makes clear what repository to look at if you read this thing as a newcomer

lib/types.ts
loweisz: Updated the commment

Jwhiles (approved):

PR #817, opened 2021-06-28
test: fix openAsset waitForSlideIn test

djagya:
# Purpose of PR

Fixes one of the slide-in asset editor tests that was failing after some internal changes in the web app. 

With those changes the callback returned from `openAsset()` with `waitForClose: true` option now triggers immediately after `clickSlideInClose()` and executed in the the same cy command context, prohibiting calling another cy command inside the promise. It caused the ["Cypress detected that you returned a promise from a command while also invoking one or more cy commands in that promise."](https://docs.cypress.io/guides/references/error-messages#Cypress-detected-that-you-returned-a-promise-from-a-command-while-also-invoking-one-or-more-cy-commands-in-that-promise) error.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/test/cypress/integration/reusable/open-asset-test.ts
+++ b/test/cypress/integration/reusable/open-asset-test.ts
@@ -59,7 +59,7 @@ export function openAssetSlideInTest(iframeSelector: string, currentEntryId: str
     })
   })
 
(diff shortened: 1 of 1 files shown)


loweisz (approved):
We should keep an eye out that this is not bringing back the flakiness fixed by https://github.com/contentful/ui-extensions-sdk/pull/812 with the boolean flag, but besides that lgtm

PR #818, opened 2021-06-29
chore: allow to run integration tests in different environments

shikaan:
## PR Checklist

- [x] Tests are passing


test/integration/contentful-client.ts
```diff
@@ -1,11 +1,89 @@
...
+
+export const getUsersByRole = async () => {
+  const space = await getCurrentSpace()
+  const organization = await client.getOrganization(space.sys.organization.sys.id)
+
+  const [users, spaceMembers, spaceRoles] = await Promise.all([
+    organization.getUsers(),
+    space.getSpaceMembers(),
+    space.getRoles(),
+  ])
```
loweisz, line 52: Could we not save some requests with using the plain client here ? 🤔

test/integration/contentful-client.ts
```diff
@@ -1,11 +1,89 @@
...
+
+  const result = {
+    admin: {} as ReturnType<typeof cleanUser>,
+    editor: {} as ReturnType<typeof cleanUser>,
+    editorMasterOnly: {} as ReturnType<typeof cleanUser>,
+  }
+
+  for (const user of users.items) {
+    // @ts-expect-error user is part of space member
```
loweisz, line 61: For me it shows there is no error anymore, can you double check ?

test/integration/contentful-client.ts
```diff
@@ -1,11 +1,89 @@
...
+    const sm = spaceMembers.items.find((m: SpaceMemberProps) => m.sys.user.sys.id === user.sys.id)
+
+    if (!sm) {
+      continue
+    }
+
+    if (sm.admin) {
+      result.admin = cleanUser(user, sm)
+    } else {
+      const roles = (sm.roles.map((i) => spaceRoles.items.find((j) => j.sys.id === i.sys.id)) ??
```
loweisz, line 71: isn't map always returning an array? So I don't think you need the `??` check here

test/integration/contentful-client.ts
shikaan: I still get the error 🤷  Have your run tsc?

test/integration/contentful-client.ts
shikaan: Yup, makes sense. I didn't want to to not touch unrelated parts, but might as well

test/integration/contentful-client.ts
shikaan: Good catch :D It's leftover before I introduced the `cotninue` statement

test/cypress/integration/reusable/check-sdk-space-methods-test.ts
```diff
@@ -1,8 +1,15 @@
...
+import { removeVariableData } from '../../utils/remove-variable-data'
+import expectedContentTypes from '../fixtures/cached-content-types'
 
 export function verifySpaceGetCachedContentTypes(iframeSelector: string) {
-  cy.getSdk(iframeSelector).then(sdk => {
-    expect(sdk.space.getCachedContentTypes()).to.deep.equal(cachedContentTypes)
+  cy.getSdk(iframeSelector).then((sdk) => {
+    const cachedContentTypes = sdk.space.getCachedContentTypes()
+
+    for (const ct of cachedContentTypes) {
```
andipaetzold, line 8: An empty array would make this test succeed. Can't we do

```typescript
expect(sdk.space.getCachedContentTypes().map(removeVariableData)).to.deep.equal(expectedContentTypes.map(removeVariableData));
```

andipaetzold (changes requested):

test/cypress/integration/fixtures/user-data.json
```diff
@@ -20,16 +20,16 @@
...
     "sys": {
       "type": "User",
-      "id": "7p08FEb4uWrGKcyt2wygpb"
+      "id": "4Qavd6zOemHk0t99Go1jHW"
     },
     "firstName": "Extensibility",
     "lastName": "Testing (Editor)",
     "email": "team-extensibility+testing-editor@contentful.com",
-    "avatarUrl": "https://www.gravatar.com/avatar/32258d2511fe07bf4c1a6b02784ac851?s=50&d=https%3A%2F%2Favatars.contentful.com%2Finitials%2Fv1%2FET%2F50",
+    "avatarUrl": "https://www.gravatar.com/avatar/32258d2511fe07bf4c1a6b02784ac851?s=50&d=https%3A%2F%2Favatars.flinkly.com%2Finitials%2Fv1%2FET%2F50",
```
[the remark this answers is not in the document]
shikaan: Good catch. This should go into the "ignored" data, else tests will fail

test/cypress/integration/reusable/check-sdk-space-methods-test.ts
shikaan: Good point. Maybe it's more readable if we iterated on the `expectedContentTypes` (which is fixed and always present), without nesting `map`s. WDYT?

andipaetzold: yes that would also work. I agree that this nesting is a little confusing. Of course we could also extract the value within `expect(...)` and `equal(...)` to it's own variable, but I actually don't care.
In case of an error many small expects probably provide a better diff.

shikaan: > In case of an error many small expects probably provide a better diff.

This was exactly the reasoning :)

andipaetzold: Then just iterate over `expectedContentTypes` 👍

test/integration/contentful-client.ts
loweisz: you are right, now I see it too 😬

loweisz (approved):
🎉🎉🎉

PR #834, opened 2021-07-02
Update canary

loweisz:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing


shikaan (approved):

andipaetzold (approved):

package.json
```diff
@@ -150,6 +150,6 @@
     ]
   },
   "dependencies": {
-    "contentful-management": "^7.23.0"
+    "contentful-management": "7.27.0"
```
andipaetzold, line 153: Why did you pin the dependency?

```suggestion
    "contentful-management": "^7.27.0"
```

andipaetzold (changes requested):
Why did you pin the dependency?

package.json
loweisz: it was pinned in master when it was under `devDependencies`: 
https://github.com/contentful/ui-extensions-sdk/blob/1b474e136ff75cda86f20036e00a31f51bef29da/package.json#L58

andipaetzold: It's fine to pin dev dependencies as people installing app-sdk do not install those. But regular dependencies should not be pinned as we fix the cma.js our users can use which will cause issues.

loweisz: Makes sense 👍

andipaetzold (approved):

PR #837, opened 2021-07-06
fix: missing space types and SerializedJSONValue

shikaan:
missing-space

Also, introduces `SerializedJSONValue` for `setValue`

### DIFF
--- a/lib/types/entities.ts
+++ b/lib/types/entities.ts
@@ -1,7 +1,6 @@
 export type {
   TagProps as Tag,
   TeamProps as Team,
-  TagVisibility,
   UserProps as User,
   AssetProps as Asset,
   TaskProps as Task,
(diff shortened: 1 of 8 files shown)


andipaetzold (approved):
So, the types failed but rollup/typescript didn't tell us?

shikaan:
> So, the types failed but rollup/typescript didn't tell us?

Yup, maybe worth adding a `tsc` in the build command

lib/types/entities.ts
```diff
@@ -17,6 +16,8 @@ export type {
   QueryOptions,
 } from 'contentful-management/types'
 
+export type { TagVisibility } from 'contentful-management/dist/typings/entities/tag'
```
andipaetzold, line 19: Should we directly export that from cma.js?

PR #863, opened 2021-07-16
fix(dialogIds):  ignore entity scoped ids from baseSDK [EXT-2878]

kdamball:


lib/types/api.types.ts
```diff
@@ -143,6 +143,8 @@ export interface AccessAPI {
   canEditAppConfig: () => Promise<boolean>
 }
 
+type EntityScopedIds = 'field' | 'entry' | 'contentType'
```
andipaetzold, line 146: Why not `EntryScopedIds`? We do not have an app location for any other entity.

andipaetzold (approved):
Just one comment about the naming

lib/types/api.types.ts
kdamball: Fair point

PR #865, opened 2021-07-16
feat: [EXT-2242] Add typings and expose metadata methods

colomolo:
# Purpose of PR
Add proper typings for a metadata of entry, return `getMetadata` and `onMetadataChange` methods at `createEntry` resulting object.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -98,6 +98,10 @@ export interface EntrySys {
   contentType: Link
 }
 
+export type Metadata = {
+  tags?: Link[]
+} | undefined
+
 /* Field API */
 export interface FieldAPI {
   /** The ID of a field is defined in an entry's content type. */
@@ -227,9 +231,9 @@ export interface EntryAPI extends TaskAPI {
    * Optional metadata on an entry
    * @deprecated
    */
-  metadata?: {
-    tags?: Link[]
-  }
+  metadata?: Metadata,
+  getMetadata: () => Metadata,
+  onMetadataChanged: (callback: (metadata: Metadata) => void) => Function,
 }
 
 /* Scheduled Actions */
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -98,9 +98,9 @@ export interface EntrySys {
   contentType: Link
 }
 
-export type Metadata = {
+export interface Metadata {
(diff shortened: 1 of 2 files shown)


lib/types.ts
```diff
@@ -98,6 +98,10 @@ export interface EntrySys {
   contentType: Link
 }
 
+export type Metadata = {
+  tags?: Link[]
+} | undefined
```
andipaetzold, line 103: I think it's better to add `undefined` to the `getMetadata` and `onMetadataChanged` types below. That way we can properly use `Metadata` as a type when we know that it is set.

lib/types.ts
```diff
@@ -227,9 +231,9 @@ export interface EntryAPI extends TaskAPI {
    * Optional metadata on an entry
    * @deprecated
    */
-  metadata?: {
-    tags?: Link[]
-  }
+  metadata?: Metadata,
+  getMetadata: () => Metadata,
+  onMetadataChanged: (callback: (metadata: Metadata) => void) => Function,
```
andipaetzold, line 236:
```suggestion
  metadata?: Metadata
  getMetadata: () => Metadata | undefined
  onMetadataChanged: (callback: (metadata?: Metadata) => void) => () => void
```

andipaetzold (changes requested):
Already looks good. Just some type changes.

lib/types.ts
colomolo: Yeah, thought about the same and fixed it.

lib/types.ts
```diff
@@ -231,9 +231,9 @@ export interface EntryAPI extends TaskAPI {
    * Optional metadata on an entry
    * @deprecated
    */
-  metadata?: Metadata,
-  getMetadata: () => Metadata,
-  onMetadataChanged: (callback: (metadata: Metadata) => void) => Function,
+  metadata?: Metadata | undefined
```
andipaetzold, line 234: nit: `metadata?: Metadata` is actually the same as `metadata?: Metadata | undefined`

andipaetzold:
Looks good - gonna approve after testing with a real app locally

andipaetzold (approved):

PR #869, opened 2021-07-19
fix: make cma an actual dependency

andipaetzold:
We are relying on `contentful-management` to be installed because our types are broken otherwise.

Peer dependencies are not automatically installed in npm 6


kdamball (approved):

loweisz (approved):

shikaan (changes requested):
As per previous conversation, I think we should not force people to depend on cma. You don't want to install all the deps cma brings along for the sake of typings, especially if you - for example - run on plain js. 

`peerDependencies` leave users with the freedom of installing the package if they need to (and advise them to do so with the warning, in old npm versions), without forcing any dependency

PR #871, opened 2021-07-20
chore: remove stale entries before each run and fix envs

andipaetzold:
# Purpose of PR

With this PR we delete stale entries from `master-test` before every run. In theory these should be deleted after the tests finish, but in some cases CI crashes and some leftovers remain in the `master-test` environment. The implementation is almost identical to removing stale environments.

The stale environment removal task had a few implementation issues:
- You can't use `.forEach` with async functions as we cannot `await` for `forEach` to finish with all promises. I therefore rewrote this to a `for ... of ...`
- We are using the plain cma client here which means that we cannot call `environment.delete()` or `entry.delete()`. I adjusted the code to use the plain client syntax for deletion.

### DIFF
--- a/test/integration/tasks/delete-stale-entries.ts
+++ b/test/integration/tasks/delete-stale-entries.ts
@@ -23,6 +23,10 @@ export default async (client = plainClient) => {
         const id = entry.sys.id
 
         try {
+          if (entry.sys.publishedVersion) {
+            await client.entry.unpublish({ entryId: id, environmentId: 'master-test' })
+          }
+
           await client.entry.delete({ entryId: id, environmentId: 'master-test' })
           console.log(`Deleted entry ${id}`)
           return id
--- a/test/integration/tasks/delete-stale-entries.ts
+++ b/test/integration/tasks/delete-stale-entries.ts
@@ -1,5 +1,6 @@
+import { EntryProps, KeyValueMap } from 'contentful-management/types'
 import { plainClient } from '../contentful-client'
-import { printStepTitle, sleep } from '../utils'
+import { printStepTitle } from '../utils'
 
 const TWO_HOURS_IN_MS = 60 * 60 * 2 * 1000
 
@@ -9,30 +10,29 @@ export default async (client = plainClient) => {
   const entryCollection = await client.entry.getMany({ environmentId: 'master-test' })
   const { items: entries } = entryCollection
 
-  const isStaleEntry = (timeStamp: string) => {
-    const entryDate = new Date(timeStamp).getTime()
+  const isStaleEntry = (entry: EntryProps<KeyValueMap>) => {
+    const entryDate = new Date(entry.sys.createdAt).getTime()
     const difference = Date.now() - entryDate
     return difference >= TWO_HOURS_IN_MS
   }
 
-  const deletedEntries: string[] = []
-  for (const entry of entries) {
-    const {
-      sys: { createdAt, id },
-    } = entry
-    if (!isStaleEntry(createdAt)) {
-      continue
-    }
+  const promiseResults = await Promise.allSettled(
+    entries
+      .filter((entry) => isStaleEntry(entry))
+      .map(async (entry) => {
+        const id = entry.sys.id
 
-    try {
-      await client.entry.delete({ entryId: id })
-      console.log(`Deleted entry ${id}`)
-      deletedEntries.push(id)
-      await sleep(200)
-    } catch (error) {
-      console.error(`Could not delete entry ${id}`)
-    }
-  }
+        try {
+          await client.entry.delete({ entryId: id, environmentId: 'master-test' })
+          console.log(`Deleted entry ${id}`)
+          return id
+        } catch (error) {
+          console.error(`Could not delete entry ${id}`)
+        }
+      })
+  )
 
-  return deletedEntries
+  return promiseResults
+    .filter((r): r is PromiseFulfilledResult<string> => r.status === 'fulfilled')
+    .map((r) => r.value)
 }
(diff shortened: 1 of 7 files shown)


test/integration/tasks/delete-stale-entries.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const entryCollection = await client.entry.getMany({ environmentId: 'master-test' })
+  const { items: entries } = entryCollection
+
+  const isStaleEntry = (timeStamp: string) => {
+    const entryDate = new Date(timeStamp).getTime()
+    const difference = Date.now() - entryDate
+    return difference >= TWO_HOURS_IN_MS
+  }
+
```
Jwhiles, line 17: You could consider using Promise.all here. I think the current approach is fine, but you called it out in the PR description. 

```
await Promise.all(entries.map(entry => {
 const {
      sys: { createdAt, id },
    } = entry

    if (!isStaleEntry(createdAt)) {
      return Promise.resolve()
    }
    try {
      await client.entry.delete({ entryId: id })
      console.log(`Deleted entry ${id}`)
      deletedEntries.push(id)
      await sleep(200)
    } catch (error) {
      console.error(`Could not delete entry ${id}`)
    }
}))
```

test/integration/tasks/delete-stale-entries.ts
```diff
@@ -0,0 +1,38 @@
...
+    } = entry
+    if (!isStaleEntry(createdAt)) {
+      continue
+    }
+
+    try {
+      await client.entry.delete({ entryId: id })
+      console.log(`Deleted entry ${id}`)
+      deletedEntries.push(id)
+      await sleep(200)
```
Jwhiles, line 31: what's the sleep for?

test/integration/tasks/delete-stale-entries.ts
andipaetzold: We want to delete things sequentially here. That's also why we have the `sleep(200)` here (I think). We just had 800 stale entries in `master-test` and otherwise very certainly run into rate limiting.

test/unit/delete-stale-entries.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+        getMany: () => Promise.resolve(entryMocks),
+        delete: () => Promise.resolve(),
+      },
+    } as unknown as PlainClientAPI
+  })
+
+  it('should delete an entry that is older than one day', async () => {
+    const expectedResult = ['b4ltnOQh2HkAFBZVMzdz7']
+    expect(await deleteStaleEntries(plainClient)).to.eql(expectedResult)
+  })
```
shikaan, line 20: I would also check we're not deleting currently used entities here

test/integration/tasks/delete-stale-entries.ts
shikaan: I agree with @Jwhiles here, maybe we can cleanup before merging this PR (e.g., run this version once, but keep this parallel for the upcoming runs. This might endup taking a long time should any problem with the test runner poop up

test/unit/delete-stale-entries.spec.ts
andipaetzold: If an entry is older than two hours, we don't use it anymore. Same as for entries. We just copy that entry from `test-base`, run the tests, done

test/integration/tasks/delete-stale-entries.ts
andipaetzold: Then I'll go for `Promise.allSettled` even though I think the risk of getting rate limited is higher than having to wait for a long time.

shikaan: Some requests would go through and in a couple of rounds we would be clean either way. So it sounds good to me. Test runs are already slow, wouldn't add much more overhead

Jwhiles (approved):
This looks good to me

test/integration/tasks/delete-stale-entries.ts
```diff
@@ -0,0 +1,42 @@
+import { EntryProps, KeyValueMap } from 'contentful-management/types'
+import { plainClient } from '../contentful-client'
+import { printStepTitle } from '../utils'
+
+const TWO_HOURS_IN_MS = 60 * 60 * 2 * 1000
+
+export default async (client = plainClient) => {
+  printStepTitle('Removing stale entries from master-test')
```
kdamball, line 8: What about `test-base`? Curious if those can lead to piling up as you had to delete nearly 1000 of them

kdamball (commented):
just one question

test/integration/tasks/delete-stale-entries.ts
andipaetzold: The entries I manually deleted were produced by an error in copy-entries.
That was resolved by https://github.com/contentful/ui-extensions-sdk/commit/4b4e073bb25628620a724b9725a986cadb526065#diff-ddc7cc8d83237e60e8073e30613fdff8e7a89cf5247c6a096d8f7a77f31955b5

kdamball (approved):

test/integration/tasks/delete-stale-environments.ts
```diff
@@ -1,38 +1,43 @@
...
 
-  const environments = await client.environment.getMany({})
-  const { items } = environments
+  const environmentCollection = await client.environment.getMany({})
+  const { items: environments } = environmentCollection
 
   // filter for relevant environments
-  const isProtected = (name: string) => name === 'master' || name.includes('test')
+  const isProtected = (environment: EnvironmentProps) =>
+    environment.name === 'master' || environment.name.includes('test')
```
shikaan, line 15: We should document this in the readme

shikaan (approved):

test/integration/tasks/delete-stale-environments.ts
andipaetzold: Added a sentence to the docs.

PR #880, opened 2021-07-23
fix: adjust comments to mention apps instead of extensions

andipaetzold:
# Purpose of PR

We call the library App SDK and apps are the latest way to extend Contentful. We should therefore also not refer to legacy extensions in our comments.

### DIFF
--- a/lib/types/api.types.ts
+++ b/lib/types/api.types.ts
@@ -66,7 +66,7 @@ export interface NotifierAPI {
 /* Location API */
 
 export interface LocationAPI {
-  /** Checks the location in which your extension is running */
+  /** Checks the location in which your app is running */
   is: (type: string) => boolean
 }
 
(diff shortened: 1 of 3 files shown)


kdamball (approved):
Always wanted this change to happen but was told no before 😅 . Happy to see it finally

loweisz:
Lgtm, I'm just wondering if we maybe could say both, "app or extension" like it was done [here](https://github.com/contentful/ui-extensions-sdk/pull/880/files#diff-ae76f63dff4e8844cc9a810b9cc63e79a33fdeecb5630ad60a50ea56c8f4b297R48)

andipaetzold:
> Lgtm, I'm just wondering if we maybe could say both, "app or extension" like it was done [here](https://github.com/contentful/ui-extensions-sdk/pull/880/files#diff-ae76f63dff4e8844cc9a810b9cc63e79a33fdeecb5630ad60a50ea56c8f4b297R48)

I'd personally wouldn't do that.
The reason why we mention both here is that `openCurrent` has a specific implementation to handle both cases. On the same API there is `openExtension` and `openCurrentApp` that only handle one case.

We renamed everything in the repo to match refert to apps (including the name). So, I think this should be fine.

loweisz:
Fair enough 👍

loweisz (approved):

fabe (approved):

PR #901, opened 2021-08-08
feat: add option to include absolute elements in auto resizer

bgutsol:
# Purpose of PR

Update the `MutationObserver` to count absolute positioned elements on the iframe.
The elements get added to the Set of elements, if they are added to the DOM with position absolute, or if an attribute on a element with position absolute is changed, used when an element is already on the DOM but with display: none for example.

They are removed when removed from the DOM, or when they are hidden, or not absolute positioned anymore.

## PR Checklist

- [ ] Tests are passing


lib/window.ts
```diff
@@ -41,7 +78,21 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
     if (height === null) {
-      height = Math.ceil(document.documentElement.getBoundingClientRect().height)
+      const documentHeight = Math.ceil(document.documentElement.getBoundingClientRect().height)
+      const fullDocumentHeight = Math.max(
+        document.body.scrollHeight,
+        document.body.offsetHeight,
+        document.body.clientHeight,
+        document.documentElement.clientHeight,
+        document.documentElement.scrollHeight,
+        document.documentElement.offsetHeight
```
andipaetzold, line 88: `scrollHeight` <= `clientHeight` <= `offsetHeight`
Why use all three of them? Any situation where this is not the case?

https://medium.com/@jbbpatel94/difference-between-offsetheight-clientheight-and-scrollheight-cfea5c196937

lib/window.ts
```diff
@@ -8,13 +8,50 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
+          const node = mutation.target
+          const computedStyle = window.getComputedStyle(node as Element)
+          if (computedStyle.position === 'absolute' && computedStyle.display !== 'none') {
+            absolutePositionedElems.add(node)
+          } else {
+            absolutePositionedElems.delete(node)
+          }
+        }
+      } else if (mutation.type === 'childList') {
+        if (mutation.addedNodes.length > 0) {
```
andipaetzold, line 38: We can remove these conditions as we are anyway iterating over the array

-> less nesting, more readability

lib/window.ts
```diff
@@ -8,13 +8,50 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
   let isAutoResizing = false
+  const absolutePositionedElems: Set<Node> = new Set()
 
   const self = { startAutoResizer, stopAutoResizer, updateHeight }
   return self
 
+  function checkAbsolutePositionedElems(mutations: Array<MutationRecord>) {
+    mutations.forEach((mutation) => {
+      if (mutation.type === 'attributes') {
+        if (mutation.target.nodeType === 1) {
```
andipaetzold, line 28: nit: Just being explicit

```suggestion
        if (mutation.target.nodeType === Node.NODE_ELEMENT) {
```

lib/window.ts
```diff
@@ -8,13 +8,50 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
+
+  const heightObserverCallback = (mutations: Array<MutationRecord>) => {
+    checkAbsolutePositionedElems(mutations)
+    autoUpdateHeight()
+  }
+
+  const observer = new MutationObserver(heightObserverCallback)
   let oldHeight: number
   let isAutoResizing = false
+  const absolutePositionedElems: Set<Node> = new Set()
```
andipaetzold, line 20: This should be `Set<Element>`, right?
We don't really want to have any other node types in here

lib/window.ts
```diff
@@ -41,7 +78,21 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
+        document.body.scrollHeight,
+        document.body.offsetHeight,
+        document.body.clientHeight,
+        document.documentElement.clientHeight,
+        document.documentElement.scrollHeight,
+        document.documentElement.offsetHeight
+      )
+
+      if (documentHeight !== fullDocumentHeight) {
+        height = absolutePositionedElems.size > 0 ? fullDocumentHeight : documentHeight
```
andipaetzold, line 92: What happens if we don't check for absolute elements here and always use `fullDocumentHeight`? Would this cause issues?

andipaetzold (changes requested):
As we are changing the way the auto resizer works, I'd consider this a breaking change. WDYT about adding a flag/option like `sdk.window.startAutoresizer({ absoluteElements: true })`? That way we do not change the behavior of existing apps.

lib/window.ts
massao: This could work for some cases, but since the `offsetHeight` would consider all elements even if not visible, so what we do is that we actually check that absolute positioned elements are also visible before adding them to the Set.

But we could always return the `offsetHeight`, in that case we wouldn't have to check manually for absolute positioned elements, WDYT??
but it could cause some issues if there are elements with transform, as `offsetHeight` gets the dimensions before transform is applied.

andipaetzold: From what I see in the F36 source, we probably don't have any issue with `transform`. But I can see some third party libraries use `transform` to position dropdowns or context menus.

Could we iterate over all absolutely positioned elements, call `getBoundingClientRect` for each and then take the largest `bottom` value to calculate the height?

andipaetzold (commented):
The types have to be updated including jsdocs: https://github.com/contentful/ui-extensions-sdk/blob/master/lib/types/window.types.ts#L5

andipaetzold:
Looks good from my side. Just want to get feedback from other's of the team before we release it.

lib/window.ts
```diff
@@ -8,25 +8,66 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
+  let checkAbsoluteElements = false
+  const absolutePositionedElems: Set<Element> = new Set()
+
+  // Start observer to get absolute elements
+  observer.observe(document.body, {
+    attributes: true,
+    childList: true,
+    subtree: true,
+    characterData: true,
+  })
```
shikaan, line 31: Should we clean up after this listener on stop?

shikaan (commented):
LGTM, just a question regarding cleaning up the listener

lib/window.ts
andipaetzold: We never stop the observer. The observer starts when initializing the sdk.

shikaan: Yeah, that was broadly m question actually. Don't we want to start/stop listening only if needed? To not introduce performance regressions in pages with tens of Apps

andipaetzold: We do start immediately in order to not miss any absolute element. If we only start listening when calling `startAutoResizer`, we will have to go through the whole dom. Also, if the user wants to use `updateHeight()` instead of the auto resizer, we would also need to go through the whole dom every time.

andipaetzold:
Quick measurement with 50 fields containing 1 menu each.

*`heightObserverCallback` takes <1ms while loading
* `checkAbsolutePositionedElems` is triggered ~6 times when opening the menu. Total time ~1ms

The number of fields does not have an impact in the mutationobserver performance as they are within different iframes

andipaetzold:
It is possible to trigger infinite loops using absolute elements. E.g. by writing

`<div style={{ position: "absolute", bottom: -5 }}>Content</div>`

That way the iframe's height is infinitely increased by 5px.

Do you think this is a blocker? @massao @shikaan

massao:
> It is possible to trigger infinite loops using absolute elements. E.g. by writing
> 
> `<div style={{ position: "absolute", bottom: -5 }}>Content</div>`
> 
> That way the iframe's height is infinitely increased by 5px.
> 
> Do you think this is a blocker? @massao @shikaan

Uhmm good point, do you think there is a use case where this could be justified??
Also any idea how we could handle this so this wouldn't cause a infinite loop??

andipaetzold:
@massao I'd try to exclude elements that could trigger infinite loops. I'll play around a bit and update this PR then

lib/window.ts
```diff
@@ -87,7 +107,7 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
       if (checkAbsoluteElements && absolutePositionedElems.size) {
         let maxHeight = documentHeight
         absolutePositionedElems.forEach((element) => {
-          maxHeight = Math.max(element.getBoundingClientRect().bottom, maxHeight)
+          if (element) maxHeight = Math.max(element.getBoundingClientRect().bottom, maxHeight)
```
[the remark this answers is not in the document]
massao: missed this 😅 
it's not needed at all as it's inside the forEach already

lib/window.ts
```diff
@@ -35,13 +96,23 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
   function updateHeight(height: number | null = null) {
     if (height === null) {
-      height = Math.ceil(document.documentElement.getBoundingClientRect().height)
+      const documentHeight = Math.ceil(document.documentElement.getBoundingClientRect().height)
+
+      // Only check for absolute elements if option is provided to startAutoResizer
+      if (checkAbsoluteElements && absolutePositionedElems.size) {
+        let maxHeight = documentHeight
+        absolutePositionedElems.forEach((element) => {
+          if (element) maxHeight = Math.max(element.getBoundingClientRect().bottom, maxHeight)
```
massao, line 110:
```suggestion
          maxHeight = Math.max(element.getBoundingClientRect().bottom, maxHeight)
```

lib/window.ts
```diff
@@ -8,25 +8,86 @@ export default function createWindow(currentWindow: Window, channel: Channel): W
...
+      !isNaN(parseInt(computedStyle.bottom)) && parseInt(computedStyle.bottom) < 0
+    if (type === 'attributes') {
+      return (
+        computedStyle.position === 'absolute' &&
+        computedStyle.display !== 'none' &&
+        !isNegativeBottom
+      )
+    } else if (type === 'childList') {
+      return computedStyle.position === 'absolute' && !isNegativeBottom
+    }
```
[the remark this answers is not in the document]
andipaetzold: I'll add this in my cleanup commit

andipaetzold:
> It is possible to trigger infinite loops using absolute elements. E.g. by writing
> 
> `<div style={{ position: "absolute", bottom: -5 }}>Content</div>`
> 
> That way the iframe's height is infinitely increased by 5px.
> 
> Do you think this is a blocker? @massao @shikaan

If we filter out elements with negative `bottom` or `transform`, elements opened by popperjs don't resize the window anymore. I therefore reverted these changes.

We should add a warning instead that infinite loops might happen if you are not careful. Or find a way to detect infinite loops which is probably quite complicated and won't work 100% of cases.

gui-santos:
@bgutsol @massao I won't review this PR hahahaha

I'm also missing you guys ❤️

lib/types/window.types.ts
```diff
@@ -3,6 +3,9 @@ export interface WindowAPI {
   updateHeight: (height?: number) => void
   /**
    * Listens for DOM changes and updates height when the size changes.
+   *
+   * When passing absoluteElements true, an infinite loop can happen if elements are always rendered bellow the height of the window.
```
andipaetzold, line 7:
```suggestion
   * When passing absoluteElements true, an infinite loop can happen if elements are always rendered below the height of the window.
```

andipaetzold (approved):

PR #913, opened 2021-08-16
Permission integration test

loweisz:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing


test/cypress/integration/fixtures/user-data.json
```diff
@@ -2,16 +2,16 @@
   "admin": {
```
shikaan, line 2: I think we can delete and ignore this file (gets regenerated on every test run)

test/cypress/integration/fixtures/ids-data.json
```diff
@@ -14,13 +14,15 @@
     "field": "title"
```
shikaan, line 14: Same as the other file

shikaan (approved):
LGTM only a minor comment

test/cypress/integration/fixtures/user-data.json
loweisz: That's a good point 🙇‍♂️

PR #972, opened 2021-09-16
feat(AccessAPI): add support for JSON patches in access sdk [EXT-3041]

kdamball:
# Purpose of PR

Add support for checking patches for all entities supported in access sdk

## PR Checklist

- [x] Tests are not required

### DIFF
--- a/lib/types/api.types.ts
+++ b/lib/types/api.types.ts
@@ -130,6 +130,11 @@ export interface SharedEditorSDK {
 type CrudAction = 'create' | 'read' | 'update' | 'delete'
 type PublishableAction = 'publish' | 'unpublish'
 type ArchiveableAction = 'archive' | 'unarchive'
+type JSONPatchItem = {
+  op: 'remove' | 'replace' | 'add'
+  path: string
+  value?: any
+}
 
 export interface AccessAPI {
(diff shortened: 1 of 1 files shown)


lib/types/api.types.ts
```diff
@@ -130,6 +130,11 @@ export interface SharedEditorSDK {
 type CrudAction = 'create' | 'read' | 'update' | 'delete'
 type PublishableAction = 'publish' | 'unpublish'
 type ArchiveableAction = 'archive' | 'unarchive'
+type PatchItem = {
```
shikaan, line 133: nit: JSONPatch is a standard, I would keep the name

shikaan (approved):
Should we do this also on v3?

lib/types/api.types.ts
```diff
@@ -146,6 +151,12 @@ export interface AccessAPI {
 
   can<T = Object>(action: ArchiveableAction, entity: 'Asset' | 'Entry' | T): Promise<boolean>
 
+  can<T = Object>(
```
andipaetzold, line 154: Can we now get rid of the generic here :P

shikaan (approved):

andipaetzold (approved):

PR #974, opened 2021-09-17
feat(AccessAPI): add support for JSON patches in access sdk [EXT-3041]

kdamball:
# Purpose of PR

Pointing to master now to release for v3

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -701,7 +701,7 @@ export interface AccessAPI {
   can<T = Object>(
     action: 'patch',
     entity: 'Asset' | 'Entry' | T,
-    patch?: JSONPatchItem[]
+    patch: JSONPatchItem[]
   ): Promise<boolean>
   /** Whether the current user can edit app config */
   canEditAppConfig: () => Promise<boolean>
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -681,6 +681,11 @@ export interface SharedEditorSDK {
 export type CrudAction = 'create' | 'read' | 'update' | 'delete'
 export type PublishableAction = 'publish' | 'unpublish'
 export type ArchiveableAction = 'archive' | 'unarchive'
+type JSONPatchItem = {
+  op: 'remove' | 'replace' | 'add'
+  path: string
+  value?: any
+}
 
 export interface AccessAPI {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "3.39.2",
+  "version": "3.39.3",
```
shikaan, line 4: I think we can let the CI do the bump here

shikaan (approved):
LGTM we might need to revert package-lock and package.json

lib/types.ts
```diff
@@ -693,6 +698,11 @@ export interface AccessAPI {
     entity: 'ContentType' | ContentType | 'Asset' | 'Entry' | T
   ): Promise<boolean>
   can<T = Object>(action: ArchiveableAction, entity: 'Asset' | 'Entry' | T): Promise<boolean>
+  can<T = Object>(
+    action: 'patch',
+    entity: 'Asset' | 'Entry' | T,
```
andipaetzold, line 703: Not quite sure about this generic. Shouldn't it just be 
```suggestion
    entity: 'Asset' | 'Entry' | string,
```
and removing the generic completely?

Or is this just to be consistent and we fix this with v4?

lib/types.ts
```diff
@@ -693,6 +698,11 @@ export interface AccessAPI {
...
+    patch?: JSONPatchItem[]
```
andipaetzold, line 704: When can `patch` be `undefined`? Doesn't that basically say "check whether you can keep the entity as it is"

lib/types.ts
kdamball: could be `Task`

lib/types.ts
kdamball: I think in this case it should be required actually

lib/types.ts
andipaetzold: Well then it's still not necessary to have a generic here. And my suggestion would still be fine.

kdamball: Per the conversation, lets keep the generic for consistency 😞

andipaetzold (approved):

PR #989, opened 2021-09-24
feat: show better warning for DATA_CLONE_ERR [EXT-2761]

alexdonets:
# Purpose of PR

Describe an issue happening when trying to pass an sdk to `openCurrentApp`

## PR Checklist

- [ ] Tests are passing


lib/channel.ts
```diff
@@ -84,15 +84,23 @@ function createSender(sourceId: string, targetWindow: Window) {
...
+        {
+          source: sourceId,
+          id: messageId,
+          method,
+          params,
+        },
+        '*'
+      )
+    } catch (e) {
+      if ((e as Error).name === 'DataCloneError') {
```
andipaetzold, line 98: nit as I just saw this PR:

Instead of assuming the `Error` type, you can simply check for `e instanceof DOMException`. Then you can also check for `DataCloneError`.

Would make my inner TypeScript monk happy :D

alexdonets: Done ;)

lib/channel.ts
```diff
@@ -84,15 +84,25 @@ function createSender(sourceId: string, targetWindow: Window) {
...
+        '*'
+      )
+    } catch (e) {
+      if (e instanceof DOMException && e.name === 'DataCloneError' && method === 'openDialog') {
+        console.warn(
+          'Error: openCurrent(App) parameters must not be functions or DOM elements. Instead, use the App SDK within the target location. Learn more about the dialog location: https://ctfl.io/app-sdk-dialog'
+        )
+      }
+
+      throw e
```
loweisz, line 104: Wouldn't that mean we show the same ("cryptic") error, but just add a additional warning in the console?
Maybe we should throw a new error with that message replacing the original one? 🤔

andipaetzold: I think it probably makes sense to fully replace that error instead.

alexdonets: The problem is that we cannot cover all cases that can trigger the error with one error message. In the edge cases where the error is triggered by something other than functions or dom elements it would be close to impossible to debug, while the default error message actually gives quite a bit of information. So the idea was to show a descriptive warning in console while still showing the original cause for the error.

loweisz: That is true, so I think we should then be not so specific that this only is true for dom elements and functions with the message either. 

Maybe we could just explain that there is something within the parameters that cannot be parsed due to the limits of the iframe communication and just add functions or dom elements as examples

alexdonets: @fabe wdyt?

fabe: > openCurrent[App] parameters could not be parsed. You likely tried to pass functions or DOM elements as a parameter. Tip: Use the App SDK directly within the dialog location.
> 
> Learn more about the dialog location: https://ctfl.io/app-sdk-dialog

Would this be more accurate?

alexdonets: This sounds good to me. @loweisz @andipaetzold wdyt?

loweisz: Makes sense 👍

fabe: Sorry, accidentally deleted this:
> openCurrent[App] parameters could not be parsed. You likely tried to pass functions or DOM elements as a parameter. Tip: Use the App SDK directly within the dialog location.
Learn more about the dialog location: https://ctfl.io/app-sdk-dialog

andipaetzold: This error message is not only about `openCurrent`, right? This may also happen with any other function we use.

fabe: This one will only be shown when opening a dialog this way (using `openCurrent` or `openCurrentApp`), for some reason we have to check for `openDialog` though. It seems like the most common scenario people see this error.

alexdonets: Yes, because we use `postMessage` with every call made through the channel. Per @fabe the error has been only/mostly reported in `openCurrentApp`/`dialog` methods, that's why we show this custom error for dialog methods only.

andipaetzold: Don't we run into exactly the same when calling `sdk.field.setValue(() => { ... })` 🤔

fabe: In that case this error wouldn't be shown (just the regular postMessage error). We can think about adding a more generic fallback error for other cases (or specific ones if we see the need). This exact scenario about the dialog has been brought up in the community a bunch of times, so a specific error will hopefully explain this better than the current one.

alexdonets: @andipaetzold @loweisz updated the text

andipaetzold (approved):

fabe (approved):

PR #997, opened 2021-09-29
fix(AccessAPI): include patch in postMessage [EXT-3041]

kdamball:
# Purpose of PR
Forgot to pass patch whenever it's available in the message

## PR Checklist

- [ ] Tests are passing


andipaetzold:
That's more a `fix:` right?

kdamball:
@andipaetzold for sure but semantic PR doesn't want to cooperate

andipaetzold (approved):

PR #1008, opened 2021-10-05
feat!: App SDK v4

shikaan:


dist/README.md
```diff
@@ -1,11 +0,0 @@
-# !! WARNING !!
-
-`cf-extension.css` file in this directory is a prebuilt committed version of legacy
```
Jwhiles, line 3: We're dropping this warning - but leaving the file? Is there a reason for that

andipaetzold: We should not drop the files, but only the final sentence

> We can remove it as soon as we reach 4.0.

And update the unpkg url maybe

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [3.34.3](https://github.com/contentful/ui-extensions-sdk/compare/v3.34.2...v3.34.3) (2021-05-18)
```
andipaetzold, line 1: The order of the changelog is wrong

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "contentful-ui-extensions-sdk",
   "description": "SDK to develop custom UI Extension for the Contentful Web App",
-  "version": "4.0.0-alpha.36",
+  "version": "3.34.3",
```
andipaetzold, line 4: Should probably be 3.42.0

.circleci/config.yml
```diff
@@ -76,6 +76,7 @@ workflows:
             branches:
               only:
                 - master
+                - canary
```
Jwhiles, line 79: Do we still want this?

Jwhiles (approved):

PR #1039, opened 2021-10-14
chore: do not assign team-extensibility as dependabot reviewer

andipaetzold:


kdamball (commented):
I would combine the two PRs into one as they are both dealing with the same issue (who gets assigned PRs). This way if there's a revert we're easily back to where we started

PR #1040, opened 2021-10-14
feat: [EXT-2755] add save and (un)publish command

loweisz:
# Purpose of PR

This adds three commands to the entry API: 

- sdk.entry.publish 
- sdk.entry.unpublish
- sdk.entry.save

Making it possible for apps having access to the EntryApi to save, publish and unpublish the entry by just calling these methods without any more arguments

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "eslint-plugin-import": "2.25.2",
     "eslint-plugin-node": "11.1.0",
     "eslint-plugin-prettier": "4.0.0",
-    "eslint-plugin-promise": "5.1.0",
+    "eslint-plugin-promise": "5.1.1",
     "eslint-plugin-react": "7.26.1",
     "eslint-plugin-standard": "5.0.0",
     "fs-extra": "10.0.0",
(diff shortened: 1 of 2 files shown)


lib/types/entry.types.ts
```diff
@@ -28,6 +28,12 @@ export interface TaskAPI {
 export interface EntryAPI extends TaskAPI {
   /** Returns sys for an entry. */
   getSys: () => ContentEntitySys
+  /** Publish the entry */
+  publish: () => void
+  /** Unpublish the entry */
+  unpublish: () => void
+  /** Saves the current changes of the entry */
+  save: () => void
```
andipaetzold, line 36: These methods return `Promise<void>`

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "contentful-ui-extensions-sdk",
+  "name": "@contentful/app-sdk",
```
andipaetzold, line 2: Please revert

andipaetzold (changes requested):

andipaetzold (approved):

PR #1173, opened 2022-02-09
fix: return same SDK instance when calling `init` multiple times [EXT-3497]

andipaetzold:
# Purpose of PR

With this PR we ensure that only a single SDK instance exists within an app.

This fixes an issue where multiple SDKs subscribe and respond to an event sent from the host application. E.g. the `sdk.app.onConfigure` hook.

This might be a breaking change if someone decides to mutate the returned SDK. With this change, the mutation would also be returned for all other `init` calls. I assume only very few people do this, so this is only a patch release.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/initialize.ts
+++ b/lib/initialize.ts
@@ -5,9 +5,10 @@ export default function createInitializer(
   currentWindow: Window,
   apiCreator: (channel: Channel, data: ConnectMessage, window: Window) => KnownSDK
 ) {
-  const connectDeferred = createDeferred()
+  const connectDeferred =
+    createDeferred<[channel: Channel, message: ConnectMessage, messageQueue: unknown[]]>()
 
-  connectDeferred.promise.then(([channel]: [Channel]) => {
+  connectDeferred.promise.then(([channel]) => {
     const { document } = currentWindow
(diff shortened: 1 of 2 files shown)


lib/initialize.ts
```diff
@@ -34,26 +36,33 @@ In order for the ui-extension-sdk to function correctly, your app needs to be ru
...
+
+          // Handle pending incoming messages.
+          // APIs are created before so handlers are already
+          // registered on the channel.
+          messageQueue.forEach((m) => {
+            // TODO Expose private handleMessage method
+            ;(channel as any)._handleMessage(m)
+          })
+
+          resolve([api, customApi])
```
loweisz, line 58: I think if you just return the value in the `then`, you don't need to create wrap it with a new promise and should get the promise automatically

loweisz (approved):
Just a nit, otherwise lgtm

Jwhiles (approved):
Looks nice!

lib/initialize.ts
andipaetzold: Good point. Should make the code less indented

PR #1194, opened 2022-02-22
docs: EXT-3234 explicitly state 5.0.0 deprecation plan

shikaan:
# Purpose of PR

Explicitly state we are going to remove Space API and its method from v5 on

### DIFF
--- a/lib/space.ts
+++ b/lib/space.ts
@@ -67,7 +67,7 @@ export default function createSpace(
       console.warn(
         `You called ${String(
           methodName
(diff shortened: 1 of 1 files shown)


andipaetzold:
Isn't it "from .... onwards"? (Not a native english speaker)

shikaan:
> Isn't it "from .... onwards"? (Not a native english speaker)

Not native either, but this seems to be correct (see [here](https://dictionary.cambridge.org/dictionary/english/from-then-on?q=from+on))

andipaetzold (approved):

andipaetzold:
Probably should merge as `fix:` so we get a release

PR #1215, opened 2022-03-16
feat: entry list result data validation and tests

bgutsol:


lib/entry-list.ts
```diff
@@ -6,6 +6,29 @@ import {
...
+export default function createEntryList(channel: Channel): EntryListAPI {
+  let _handler: OnEntryListUpdatedHandler | null = null
+
+  channel.addHandler(
+    'entryListUpdated',
+    ({ msgId, props }: { msgId: string; props: OnEntryListUpdatedHandlerProps }) => {
+      return runHandler(_handler, props).then((result: OnEntryListUpdatedHandlerReturn) => {
+        return channel.send('entryListResult', { msgId, result })
+      })
+    }
```
andipaetzold, line 18: Nit: vs code proposes to convert this to an async function, maybe we can do that to get some newer-looking code

lib/entry-list.ts
```diff
@@ -6,6 +6,29 @@ import {
...
+      if (!isFunction(handler)) {
+        throw new Error('OnEntryListUpdated handler must be a function')
+      } else {
+        _handler = handler
+      }
+    },
+  }
+}
+
 const isFunction = (f: any) => typeof f === 'function'
```
andipaetzold, line 32: Can we extract this as it's also used in `app.ts`?
Or maybe inline the function directly as it's so trivial?

lib/entry-list.ts
```diff
@@ -19,32 +42,39 @@ const runHandler = async (
...
+    return validateResult(result)
   } catch (error) {
     console.error(error)
     return false
   }
 }
 
-export default function createEntryList(channel: Channel): EntryListAPI {
-  let _handler: OnEntryListUpdatedHandler | null = null
+const validateResult = (result: OnEntryListUpdatedHandlerReturn) => {
```
andipaetzold, line 53: It feels strange that the method returns a value which is always identical to the input or throws an error.
I'd make this an assert method that never returns anything but throws if something is wrong.

Now it feels like it's also doing some data cleanup or similar

lib/entry-list.ts
```diff
@@ -19,32 +42,39 @@ const runHandler = async (
...
-    onEntryListUpdated(handler: OnEntryListUpdatedHandler) {
-      if (!isFunction(handler)) {
-        throw new Error('OnEntryListUpdated handler must be a function')
-      } else {
-        _handler = handler
-      }
-    },
-  }
+    if (!schema[key](data[key])) {
+      throw new Error(`EntryListResult data is invalid. Invalid value of key "${key}"`)
```
andipaetzold, line 75:
```suggestion
      throw new Error(`EntryListResult data is invalid. Invalid value of key "${key}".`)
```

test/unit/entry-list.spec.ts
```diff
@@ -80,6 +80,40 @@ describe('createEntryList()', () => {
 
       it('returns false when a promise rejects', () =>
         test(() => Promise.reject(new Error()), false))
+
+      it('returns false if the result data hasn`t valid key', () =>
```
andipaetzold, line 84:
```suggestion
      it('returns false if the result data has invalid key', () =>
```

test/unit/entry-list.spec.ts
```diff
@@ -80,6 +80,40 @@ describe('createEntryList()', () => {
...
+            data: {
+              wrongKey: {
+                entryId: 'value',
+              },
+            } as any,
+          }),
+          false
+        ))
+
+      it('returns false if the result data hasn`t valid values type', () =>
```
andipaetzold, line 96:
```suggestion
      it('returns false if the result data has an invalid value type', () =>
```

andipaetzold (changes requested):
Looks good. Just some nits.

bgutsol:
> Looks good. Just some nits.

@andipaetzold thanks a lot for the feedback! 
Please check one more time to be sure I understood you right.

andipaetzold (approved):

PR #1263, opened 2022-05-06
fix: Use ResizeObserver to track iframe body size changes [EXT-3733]

colomolo:
# Purpose of PR
Replace `resize` event with ResizeObserver. Resize event does not trigger when the contents of iframe change their dimensions.


andipaetzold (approved):
Looks good

PR #1294, opened 2022-06-15
feat: added worflowdefinition type to exported entities and to api.types

mdmjg:
# Purpose of PR
In order to call `access.can` inside the workflows marketplace app, we need to update the apps SDK so that `WorkflowDefinition` is typed as a `PatchEntity`


Lelith (approved):
Looks good to me!

PR #1308, opened 2022-06-24
fix: make init a noop for non-browser environments [EXT-3776]

andipaetzold:
# Problem

Currently, when importing the App SDK in a non-browser environment where `window` and `document are not available, an error is thrown

```
Uncaught ReferenceError: window is not defined
```

This causes issues when building apps with SSR.

# Solution

Instead of passing around `window`, `globalThis` is used now. `globalThis` is available in all environments. Once checked whether `window` and `document` is available, `globalThis` equals `window`.

Passing around `globalThis` also improves type safety as it has the `MutationObserver` and `ResizeObserver`.


kdamball (approved):

lib/initialize.ts
```diff
@@ -2,21 +2,29 @@ import { ConnectMessage, KnownSDK } from './types'
...
-  apiCreator: (channel: Channel, data: ConnectMessage, window: Window) => KnownSDK
+  currentGlobal: typeof globalThis,
+  apiCreator: (channel: Channel, data: ConnectMessage, currentGlobal: typeof globalThis) => KnownSDK
 ) {
+  if (
+    typeof currentGlobal.window === 'undefined' ||
+    typeof currentGlobal.document === 'undefined'
+  ) {
+    // make `init` a noop if window or document is not available
+    return () => {}
```
loweisz, line 13: Shall we maybe print a warning to the console in this case?

andipaetzold: I thought about it, but we'd print for every render when using SSR. That would be quite spammy for an expected behavior

andipaetzold:
Merging on Monday

PR #1314, opened 2022-06-30
feat: export space entity type

veu:
# Purpose of PR

Export space entity for use in field editors

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/types/entities.ts
+++ b/lib/types/entities.ts
@@ -11,6 +11,7 @@ export type {
   RoleProps as Role,
   ScheduledActionProps as ScheduledAction,
   SpaceMembershipProps as SpaceMembership,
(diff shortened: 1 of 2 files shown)


andipaetzold:
Which API method returns/accepts `Space`? What do you need this type for?

veu:
@andipaetzold The WrappedEntryCard allows showing the Content Type name and I want to add an equivalent for the space. This will require the space type to be exported here which will be [used like the Content Type](https://github.com/contentful/field-editors/blob/master/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx#L33).

andipaetzold:
@veu Where does that Space data come from? Where is that object created?

PR #1319, opened 2022-07-12
feat: `init` can be called at any time [EXT-3817]

andipaetzold:
When `init` is called and no `connect` message has been received at that point in time, an `init` message is sent to the parent window.


package.json
```diff
@@ -42,6 +42,7 @@
     "@types/cross-spawn": "6.0.2",
     "@types/fs-extra": "9.0.13",
     "@types/jsdom": "16.2.14",
+    "@types/mocha": "9.1.1",
```
loweisz, line 45: why is that needed? 🤔

andipaetzold: Because we currently don't have any types and the whole editor is red.
This is not directly related to the ticket, but made my life easier.

lib/initialize.ts
```diff
@@ -36,34 +36,25 @@ export default function createInitializer(
...
 
         let customApi
         if (typeof makeCustomApi === 'function') {
           customApi = makeCustomApi(channel, params)
         }
 
-        // Handle pending incoming messages.
-        // APIs are created before so handlers are already
-        // registered on the channel.
-        messageQueue.forEach((m) => {
```
shikaan, line 60: This might be the place for the tracking, for instance

shikaan (approved):
The change looks good to me (thanks for the cleaning up too). I was just wondering if it's worth tracking this potentially breaking change

andipaetzold:
This PR has been released as [4.10.0-alpha.1](https://www.npmjs.com/package/@contentful/app-sdk/v/4.10.0-alpha.1).

PR #1330, opened 2022-07-25
Feat/get tasks query params [ZEND-2486]

ronaldronson:
# Purpose of PR

Bump contentful-sdk
Update types of the entry.getTasks method

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/types/entry.types.ts
+++ b/lib/types/entry.types.ts
@@ -12,11 +12,17 @@ export interface TaskInputData {
   dueDate?: string
 }
 
+export interface BasicQueryOptions {
+  skip?: number
+  limit?: number
+  [key: string]: any
+}
+
 /** Allows accessing the Task API for the current entry. */
 export interface TaskAPI {
   getTask(id: string): Promise<Task>
 
-  getTasks(): Promise<CollectionResponse<Task>>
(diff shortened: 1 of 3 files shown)


lib/types/entry.types.ts
```diff
@@ -12,11 +12,17 @@ export interface TaskInputData {
   dueDate?: string
 }
 
+export interface BasicQueryOptions {
```
Chaoste, line 15: Since we also allow `order`, we can also just use `SearchQuery` from [here](https://github.com/contentful/ui-extensions-sdk/blob/12cbe1a8f27e56dfb3a0764ee89c3377be3ab6f5/lib/types/utils.ts#L64), right?

Chaoste (approved):

lib/types/entry.types.ts
```diff
@@ -16,7 +16,7 @@ export interface TaskInputData {
 export interface TaskAPI {
   getTask(id: string): Promise<Task>
 
-  getTasks(): Promise<CollectionResponse<Task>>
+  getTasks(query: SearchQuery): Promise<CollectionResponse<Task>>
```
andipaetzold, line 19: The query param should be optional. This is a breaking change!

PR #1322, opened 2022-07-18
feat: add home location [EXT-3684]

martin3walker:
# Purpose of PR
We need to add a home location to the app sdk in order to support the new custom environment home functionality being added to the web app.

## PR Checklist

- [ ] Tests are added/updated/not required
- [ ] Tests are passing
- [ ] Typescript typings are added/updated/not required

PR #1388, opened 2022-10-24
feat: add home location [EXT-3684]

andipaetzold:
Same as #1322 but merging to `master`.

`canary` is not clean, so this is a recreation of the previous PR to make a nice release.


colomolo (approved):
💎

PR #1442, opened 2022-12-28
refactor: type Signal & MemoizedSignal

andipaetzold:
# Purpose of PR

Some (4) `MemoizedSignal`s are not initialized correctly and therefore return `undefined` when subscribing to their value. This PR adds proper types to `Signal` and `MemoizedSignal`, so this mistake is more obvious and won't happen in the future. This does not fix the initial values of these four signals.

## PR Checklist

- [x] Tests are passing


lib/channel.ts
```diff
@@ -54,9 +54,9 @@ export class Channel {
     this._send(method, params)
   }
 
-  addHandler(method: string, handler: Function) {
+  addHandler(method: string, handler: (...args: any[]) => void) {
```
vikaskumr, line 57: Just a quick question, why can't we use` unknow[]` instead of `any[]`, I might be missing the context here.

lib/field-locale.ts
```diff
@@ -12,9 +12,9 @@ export default class FieldLocale implements FieldAPI {
   items?: Items
   private _value: any
 
-  private _valueSignal: MemoizedSignal
-  private _isDisabledSignal: MemoizedSignal
-  private _schemaErrorsChangedSignal: MemoizedSignal
+  private _valueSignal: MemoizedSignal<[any]>
```
vikaskumr, line 15: same here.

andipaetzold: `any` here is fine. That's what we use for field values within the App SDK.

lib/signal.ts
```diff
@@ -1,14 +1,16 @@
-class Signal {
+type Listener<T extends any[]> = (...args: T) => void
```
vikaskumr, line 1: may be we can also change this to `T extends unknown[]` 🤔

vikaskumr (approved):
:shipit:

PR #1444, opened 2022-12-28
feat: initial values & getters for field/editor API

andipaetzold:
# Purpose of PR

When passing a callback to the following methods, it was called
with `undefined` instead of it's initial value
- sdk.field.onIsDisabledChanged
- sdk.field.onSchemaErrorsChanged
- sdk.editor.onLocaleSettingsChanged
- sdk.editor.onShowDisabledFieldsChanged

This PR now accepts initial values sent via the connect message and returns those instead of `undefined`

## PR Checklist

- [x] Tests are passing


lib/types/field-locale.types.ts
```diff
@@ -29,12 +29,34 @@ export interface FieldAPI {
...
    **/
   onValueChanged: (callback: (value: any) => void) => () => void
-  /** Calls the callback when the disabled status of the field changes.
-   *  the returned function can be called to remove the handler function
-   **/
+
+  /**
+   * Returns whether the current field is disabled
+   *
+   * The disabled state can change. To always work with the latest settings, use `onIsDisabledChanged`.
```
andipaetzold, line 36:
```suggestion
   * The disabled state can change. To always work with the latest state, use `onIsDisabledChanged`.
```

lib/field-locale.ts
```diff
@@ -82,10 +85,18 @@ export default class FieldLocale implements FieldAPI {
     return this._valueSignal.attach(handler)
   }
 
+  getIsDisabled(): boolean {
+    return this._isDisabledSignal.getMemoizedArgs()[0]
```
kdamball, line 89: Will this always be the same value? i.e: always called with initial value

lib/field.ts
```diff
@@ -46,12 +48,12 @@ export default class Field implements EntryFieldAPI {
...
-  setValue(value: any, locale?: string) {
-    return this._getFieldLocale(locale).setValue(value)
+  async setValue(value: any, locale?: string) {
+    return await this._getFieldLocale(locale).setValue(value)
   }
 
-  removeValue(locale?: string) {
-    return this.setValue(undefined, locale)
+  async removeValue(locale?: string) {
+    await this.setValue(undefined, locale)
```
kdamball, line 56: did you mean `return await this.setValue`

lib/api.ts
```diff
@@ -99,9 +99,9 @@ function makeWindowAPI(channel: Channel, _data: any, currentGlobal: typeof globa
 }
 
 function makeEditorAPI(channel: Channel, data: any) {
-  const { editorInterface } = data
+  const { editorInterface, editor } = data
   return {
-    editor: createEditor(channel, editorInterface),
```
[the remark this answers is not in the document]
andipaetzold: Not sure why I did that tbh... we got typescript, so we should be safe

lib/field-locale.ts
andipaetzold: No, `getMemoizedArgs()` always returns the most recent arguments. The reason for `[0]` is that the values are stored as an array, but we only care about the first.

lib/field.ts
andipaetzold: No. This is `removeValue`, so we don't want to return anything -> return `undefined`

lib/field-locale.ts
kdamball: Ah - we're prepending instead of appending the last values

lib/field.ts
kdamball: Aren't we changing the signature then?

lib/field-locale.ts
andipaetzold: No, checkout the implementation of MemoizedSignal: https://github.com/contentful/ui-extensions-sdk/blob/master/lib/signal.ts#L24. The signal always only keeps the last value. But the value is always an array because it expects to store the list of arguments of a function call. In our case we only care about the first argument of the function call.

lib/field.ts
andipaetzold: The API stays the same and the behavior is as well.
Only the types started complaining, so I changed things a bit. `setValue` returns `SerializedJSONValue | undefined` which does not match the `undefined`/`void` return value of `removeValue`. So, instead of forwarding a return value that will always be `undefined`, I now don't return anything as that's `void` by default.

kdamball (approved):

lib/types/api.types.ts
```diff
@@ -125,14 +125,47 @@ export interface EditorLocaleSettings {
...
+
+    /**
+     * Returns whether or not disabled fields are displayed
+     *
+     * This setting can change. To always work with the latest settings, use `onShowDisabledFieldsChanged`.
+     */
+    getShowDisabledFields(): boolean
+
+    /**
+     * Subscribes to changes of whether or not disabled fields are displayed
```
DanweDE, line 154: **Just an FYI:** The editor feature was rebranded to "Show hidden fields" recently (see [changelog entry](https://www.contentful.com/developers/changelog/#hidden-entry-editor-fields))

Would it make sense to reflect that also in the SDK and deprecate the old functions to avoid confusion?

andipaetzold: I already mixed the two types of "disabled" during the implementation of this PR. So, we will deprecated `onIsDisabledChanged` and ho for "hidden"

lib/types/api.types.ts
```diff
@@ -274,6 +316,10 @@ export interface ConnectMessage {
   ids: IdsAPI
   contentType: ContentTypeAPI
   editorInterface?: EditorInterface
+  editor?: {
+    localeSettings: EditorLocaleSettings
+    showDisabledFields: boolean
```
andipaetzold, line 321: @kdamball Probably we should also change the prop name of the connect message, right? That way the connect message would have a different name than the event (`showDisabledFieldsChanged`).

kdamball: I would change everything to match this new property name and remove any doubt

PR #1505, opened 2023-04-04
fix: [] Deprecate XXXExtensionSDK types in favour of XXXAppSDK

loweisz:
# Purpose of PR

Adding new duplicate types XXXAppSDK as we mainly support apps instead of extensions. 
This will also deprecate XXXExtensionSDK to be removed in the next major version.

## PR Checklist

- [ ] Tests are passing


lib/types/api.types.ts
```diff
@@ -295,6 +304,90 @@ export type KnownSDK =
   | AppExtensionSDK
   | HomeExtensionSDK
 
+export interface BaseAppSDK {
+  /** @deprecated since version 4.0.0 consider using the CMA instead
```
kdamball, line 308: I wouldn't put spaceAPI anymore for anyone using AppSDK, wdyt?

loweisz: wouldn't that make it a breaking change? Or is it fine because it is still possible to use the BaseExtensionSDK? 🤔

andipaetzold (commented):
What's the reason for duplicating the types rather than renaming and exporting an alias type? Now, if v5 takes more months and we make changes to the API, we have to make every change twice.

lib/types/api.types.ts
```diff
@@ -211,6 +211,7 @@ export interface AccessAPI {
 
 type EntryScopedIds = 'field' | 'entry' | 'contentType'
 
+/** @deprecated consider using BaseAppSDK */
```
andipaetzold, line 214: nit: Can we point to the correct type using some jsdoc magic?

lib/types/api.types.ts
```diff
@@ -271,29 +271,56 @@ export type DialogExtensionSDK = Omit<BaseExtensionSDK, 'ids'> & {
...
+export type PageExtensionSDK = PageAppSDK
+
+/** @deprecated consider using {@link HomeAppSDK} */
+export type HomeExtensionSDK = HomeAppSDK
+
+/** @deprecated consider using {@link ConfigAppSDK} */
+export type AppExtensionSDK = ConfigAppSDK
+
+/** @deprecated consider using {@link KnownAppSDK} */
+export type KnownSDK = KnownAppSDK
```
andipaetzold, line 323: `KnownSDK` is still used in `api.ts`, `index.ts` and `initialize.ts`

andipaetzold (changes requested):
Please merge as `fix:` to release a new version. And maybe a more actionable name: "Deprecate XXXExtensionSDK types in favor of XXXAppSDK" or similar

lib/types/api.types.ts
loweisz: Good catch 👍

lib/types/api.types.ts
loweisz: I agree that would make sense, but with the aliasing now I don't think that is so straight forward anymore 😬

andipaetzold (approved):

PR #1529, opened 2023-05-25
feat: [EXT-3648][EXT-4380] Improve sdk.field types

vikaskumr:
# Purpose of PR
Narrow `field` types, particularly properties `type`, `linkType`, `items`.

# Result
image


lib/field-locale.ts
```diff
@@ -1,15 +1,16 @@
...
 
 export default class FieldLocale implements FieldAPI {
   id: string
   locale: string
-  type: string
+  type: FieldLocaleType
   required: boolean
   validations: any[]
-  items?: Items
+  items: never
```
colomolo, line 13: Why it's `never`?

lib/field-locale.ts
```diff
@@ -1,15 +1,16 @@
...
-import { FieldAPI, FieldInfo, Items, SerializedJSONValue } from './types'
+import { FieldAPI, FieldInfo, SerializedJSONValue } from './types'
+import { FieldLocaleType } from './types/utils'
 import { ValidationError } from './types/validation-error'
 
 export default class FieldLocale implements FieldAPI {
   id: string
   locale: string
-  type: string
+  type: FieldLocaleType
```
colomolo, line 10: Naming could be better. This is a field type, right? And no need to use the word 'type' in the name of type.

lib/field-locale.ts
```diff
@@ -1,15 +1,16 @@
 import { Channel } from './channel'
 import { MemoizedSignal } from './signal'
-import { FieldAPI, FieldInfo, Items, SerializedJSONValue } from './types'
+import { FieldAPI, FieldInfo, SerializedJSONValue } from './types'
+import { FieldLocaleType } from './types/utils'
```
colomolo, line 4: Feels like this should be in './types'

lib/types/utils.ts
```diff
@@ -60,12 +60,18 @@ export interface EntrySys extends ContentEntitySys {
   automationTags: Link<'Tag'>[]
 }
 
+type FieldType = 'Symbol' | 'Number' | 'Array' | 'Reference'
 export interface Items {
-  type: string
-  linkType?: string
+  type: FieldType
```
colomolo, line 65: Can Items be of type 'Array'? Worth checking.

lib/types/utils.ts
```diff
@@ -60,12 +60,18 @@ export interface EntrySys extends ContentEntitySys {
...
+  linkType?: FieldType extends 'Reference' ? 'Entry' | 'Asset' : never
   validations?: ContentTypeFieldValidation[]
 }
 
+export type FieldLocaleType =
+  | { type: 'Symbol' }
+  | { type: 'Number' }
+  | { type: 'Array'; items: Items }
```
colomolo, line 73: I think 'Reference' should be also here?

colomolo (commented):
In general, I would take another approach. We have three distinctive types with common properties and individual properties. So we can use discriminated unions to define them (see 'Discriminated union' [here](https://blog.logrocket.com/understanding-discriminated-union-intersection-types-typescript/)):
```typescript
interface FieldBase {
    id: string
    name: string
    locale: string
    required: boolean
    validations: ContentTypeFieldValidation[]
    value: any
    isDisabled: boolean
    schemaErrors: ValidationError[]
}

interface SymbolField extends FieldBase {
    type: 'Symbol'
}

interface NumberField extends FieldBase {
    type: 'Number'
}

interface ArrayField extends FieldBase {
    type: 'Array'
    items: Items
}

interface ReferenceField extends FieldBase {
    type: 'Reference'
    linkType: 'Entry' | 'Asset'
}

type FieldInfo = SymbolField | NumberField | ArrayField | ReferenceField
```

Not sure if this will cause us problems down the line, but should be something along these lines.

lib/types/utils.ts
```diff
@@ -60,11 +60,23 @@ export interface EntrySys extends ContentEntitySys {
...
+  type: FieldType
   validations?: ContentTypeFieldValidation[]
 }
+export interface ReferenceItems extends ItemsBase {
+  type: 'Reference'
+  linkType: 'Entry' | 'Asset'
+}
+
+export interface PrimitiveItems extends ItemsBase {
+  type: 'Symbol' | 'Number' | 'Array'
```
colomolo, line 74: Not sure ifan item can be of 'Array' type. Needs check.

package-lock.json
andipaetzold, line 1: Please revert

andipaetzold (changes requested):
You have now properly typed the internal connect message. Any reason for not typing the external api (FieldAPI & EntryFieldAPI) as well?

colomolo:
Maybe not the best naming: `BasicEntryFieldInfo`, `ExhaustiveEntryFieldAPI`, suggestions are more than welcome.

lib/field-locale.ts
```diff
@@ -23,8 +26,12 @@ export default class FieldLocale implements FieldAPI {
     this.type = info.type
     this.required = info.required
     this.validations = info.validations
-    this.items = info.items
-
+    if (info.type === 'Array') {
```
andipaetzold, line 29: Does the condition change anything? If the type is not array, `items` is anyway initialized with `undefined`. So, this only prevents issues if we send an invalid connect message, right?

lib/types/index.ts
```diff
@@ -93,6 +93,7 @@ export type {
   EntrySys,
   ContentEntitySys,
   ContentEntityType,
+  FieldType,
```
andipaetzold, line 96: Also export `FieldLinkType`?

andipaetzold (approved):
Must be merged as `feat:` or `fix:` to trigger a new release

About the name: I have no better name myself, but as we are not exporting that type this should not be a blocker. Maybe a native speaker has a better name.

lib/field-locale.ts
colomolo: Typescript doesn't like if we do it in a straightforward way:
```js
this.items = info.items
this.linkType = info.linkType
```
> Property 'items' does not exist on type 'BasicFieldInfo'.ts(2339)

We can do it like this:
```js
this.items = (info as ArrayFieldAPI).items
```
But I wouldn't export `ArrayFieldAPI` just for this one case.

So yeah, just an extra check.

colomolo: Btw, we can delete `items` and `linkType` if `type` is not an Array or Link 🤔 Wdyt?

lib/types/index.ts
colomolo: Do you think we should expose it?

andipaetzold: We export types (`EntryFieldAPI`) that use this type, so yes.

colomolo: Done

mgoudy91 (approved):
Still coming up to speed but having seen the previous issues with typings I think this is a clean solution.

Also, I think the names are fine and fit the rest of the implementation well enough and wouldn't worry about renaming.

lib/types/utils.ts
```diff
@@ -60,12 +60,35 @@ export interface EntrySys extends ContentEntitySys {
...
+  | 'Symbol'
+  | 'Text'
+  | 'RichText'
+  | 'Number'
+  | 'Integer'
+  | 'Array'
+  | 'Link'
+  | 'Object'
+  | 'Date'
+  | 'Location'
```
gabiayako, line 73: Isn't the type Boolean necessary here?

colomolo: Good catch! I will add it.

colomolo: https://github.com/contentful/ui-extensions-sdk/pull/1541

PR #1531, opened 2023-05-30
feat: add field linkType prop

IgnacioNMiranda:
# Purpose of PR

Expose field name in related APIs for convenience.
For instance, when using the React hook `useSDK` and trying to access the field data, we cannot differentiate between a One Reference field and a One Media field.

image


colomolo:
@IgnacioNMiranda thanks for contribution. We're working on this part and will merge it as soon as other parts will be ready.

PR #1536, opened 2023-06-07
feat: expose cma in the sdk [EXT-3599]

rusticpenguin:
# Purpose of PR
In order to use CMA.js within your app, you have to install the dependency and copy a few lines to initialize the CMA client using the CMA adapter: [GitHub - contentful/contentful-management.js: JavaScript library for Contentful's Management API (node & browser)](https://github.com/contentful/contentful-management.js/#app-framework) 

This is complicated and confusing for new users if you are just getting started building apps. To improve the DX, the CMA should be exposed via sdk.cma

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "rollup-plugin-terser": "7.0.2",
     "rollup-plugin-typescript2": "0.34.1",
     "semantic-release": "19.0.5",
-    "sinon": "15.1.2",
+    "sinon": "15.2.0",
     "sinon-chai": "3.7.0",
     "ts-mocha": "10.0.0",
     "tslib": "2.5.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,7 @@
         "rollup-plugin-terser": "7.0.2",
         "rollup-plugin-typescript2": "0.34.1",
         "semantic-release": "19.0.5",
-        "sinon": "15.1.2",
+        "sinon": "15.2.0",
         "sinon-chai": "3.7.0",
         "ts-mocha": "10.0.0",
         "tslib": "2.5.3",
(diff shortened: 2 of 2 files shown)


rusticpenguin:
Do we want to deprecate the import/usage of the cmaAdapter, since this PR will allow the developer to access the CMA?

lib/api.ts
```diff
@@ -60,9 +61,10 @@ function makeSharedAPI(channel: Channel, data: ConnectMessage): BaseAppSDK {
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
   return {
+    cma: createCMAClient(ids, channel),
     cmaAdapter: createAdapter(channel),
     location: {
-      is: (tested) => currentLocation === tested,
+      is: (tested: any) => currentLocation === tested,
```
andipaetzold, line 67:
```suggestion
      is: (tested: string) => currentLocation === tested,
```

andipaetzold (changes requested):
- People will still need `cmaAdapter` for the nested cma client or for different defaults. I am not sure how many people would be affected by removing `cmaAdapter` because they cannot migrate to `sdk.cma`
- `contentful-management` is currently a peer dependency. With this change, it must be a direct dependency.

andipaetzold (approved):

lib/api.ts
```diff
@@ -60,9 +61,10 @@ function makeSharedAPI(channel: Channel, data: ConnectMessage): BaseAppSDK {
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
   return {
+    cma: createCMAClient(ids, channel),
     cmaAdapter: createAdapter(channel),
     location: {
-      is: (tested) => currentLocation === tested,
+      is: (tested: string) => currentLocation === tested,
```
ghepting, line 67: lol i almost fixed this too 😄

ghepting: i also almost added a `get` method here that _just returns the freaking string_!

ghepting (approved):

lib/api.ts
rusticpenguin: That's what I was thinking too! Don't make me guess what location I'm in haha

package.json
```diff
@@ -33,6 +33,9 @@
     "prepare": "husky install",
     "lint-staged": "lint-staged"
   },
+  "dependencies": {
+    "contentful-management": ">=7.30.0"
```
andipaetzold, line 37: We can remove CMA.js from the dev deps.

colomolo: Done

PR #1544, opened 2023-06-22
chore: update to Node 18

andipaetzold:
# Purpose of PR

Update to node 18 which required dependant chore tasks:
* update eslint because of peer dep issues
* update semantic-release because v19 is not compatible with Node 18
* cleanup dependabot.yml


colomolo (approved):
👍

PR #1560, opened 2023-06-29
feat: [EXT-3599] Bundle contentful-management with app-sdk. (Third attempt) 

colomolo:
# Purpose of PR
Including contentful-management dependency in a bundle to support legacy UI Extensions.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -36,9 +36,6 @@ module.exports = [
       format: 'umd',
       name: 'contentfulExtension',
       footer: 'globalThis.contentfulApp = globalThis.contentfulExtension;',
-      globals: {
-        'contentful-management': 'contentfulManagement',
-      },
     },
     [nodeResolve({ browser: true }), commonjs()]
   ),
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,5 +1,6 @@
 const typescript = require('rollup-plugin-typescript2')
 const { terser } = require('rollup-plugin-terser')
+const { nodeResolve } = require('@rollup/plugin-node-resolve')
 
 const pkg = require('./package.json')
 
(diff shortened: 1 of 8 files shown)


rollup.config.js
```diff
@@ -1,27 +1,42 @@
...
 
 module.exports = [
   makeConfigForOutput({
     file: pkg.main,
     format: 'umd',
     name: 'contentfulExtension',
-    footer: 'globalThis.contentfulApp = globalThis.contentfulExtension;'
-  })
+    footer: 'globalThis.contentfulApp = globalThis.contentfulExtension;',
+    globals: {
```
andipaetzold, line 29: Are these globals required?

colomolo: Yes, Rollup requires them when building.

colomolo: Docs are so confusing.

As I understood, external does not include dependency into bundle:
https://rollupjs.org/configuration-options/#external
> an Array of module IDs, or regular expressions to match module IDs, that should remain external to the bundle. Can also be just a single ID or regular expression. The matched IDs should be either:

And on `globals`:
https://rollupjs.org/configuration-options/#output-globals
> Specifies id: variableName pairs necessary for external imports in umd/iife bundles.

I will try with external.

colomolo: Result:
```
./lib/index.ts → dist/cf-extension-api.js...
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
contentful-management (guessing 'contentfulManagement')
created dist/cf-extension-api.js in 2.7s

./lib/index.ts → dist/cf-extension-api.bundled.js...
created dist/cf-extension-api.bundled.js in 5.2s
```

andipaetzold: So, `globals` is used to tell rollup that a specific dependency is stored at `globalThis.xxx`. E.g. if you load `jQuery` it assigns itself to `globalThis.$`. `external` adds a `require('...')` statement to the build which is used by the consumer of App SDK to import contentful management.

andipaetzold (approved):

PR #1565, opened 2023-06-30
fix: [EXT-3666] Remove duplicating event dispatch

colomolo:
# Purpose of PR
We're effectively dispatching change event twice – once for `setValue` then again on value changed. Removing `setValue` dispatch.


andipaetzold:
This PR means that we'll send the dispatch later which might change the behavior of apps if we expect the event to be sent immediately. This might cause issues if you have tight coupling between the sdk and the UI 🤔 

Wdyt about changing the Signal to only emit an event if something actually changed?

colomolo:
> This PR means that we'll send the dispatch later which might change the behavior of apps if we expect the event to be sent immediately. This might cause issues if you have tight coupling between the sdk and the UI 🤔

It's also strange that `removeValue` doesn't have this dispatch, but de-facto it also modifies value. So this is inconsistent.

> Wdyt about changing the Signal to only emit an event if something actually changed?

Idk, that makes sense to me, but what if somebody wants to track the change event itself, not the actual change? Feels like we'll break something for someone here in one way or another 😀

ghepting (approved):

colomolo:
**Summary**
We decided to remove this duplication and also add check for an actual value change (in a separate PR) as it is the closest to expected and conventional behaviour.

PR #1609, opened 2023-08-01
chore: Update CODEOWNERS []

ghepting:
# Purpose of PR

update CODEOWNERS


CODEOWNERS
andipaetzold, line 1: seems like we have two codeowners files. We should remove one of them. I'd the one in the root directory.

andipaetzold (approved):

PR #2064, opened 2024-11-05
fix: remove .npmrc at root to fix publish registry [EXT-5905]

whitelisab:
# Purpose of PR

Removes `.npmrc` at the root to see if the publish step in CI will respect the registry being set [here](https://github.com/contentful/ui-extensions-sdk/blob/main/.circleci/config.yml#L46).

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@
 # editors
 .vscode/
 .idea/
+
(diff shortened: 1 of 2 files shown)


BobHemphill76:
would it be possible to add .npmrc to .gitignore now too?

BobHemphill76 (approved):

PR #2078, opened 2024-11-25
feat: [] add taxonomy entities in CMA client

maxcheremisin:
# Purpose of PR

Adds taxonomy entities types in `CMAClient` interface so it properly reflects CMA methods that should be accessible in `BaseAppSDK['cma']`.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/types/cmaClient.types.ts
+++ b/lib/types/cmaClient.types.ts
@@ -14,6 +14,8 @@ export type CMAClient = {
   bulkAction: PlainClientAPI['bulkAction']
   comment: PlainClientAPI['comment']
   contentType: PlainClientAPI['contentType']
+  concept: PlainClientAPI['concept']
+  conceptScheme: PlainClientAPI['conceptScheme']
   editorInterface: PlainClientAPI['editorInterface']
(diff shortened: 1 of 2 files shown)


mgoudy91 (dismissed):
Looks great to me. 

Quick nit if you don't mind: looks like the cma client types are alphabetized, might as well maintain that convention

mgoudy91 (approved):

cgcompassion:
@maxcheremisin @mgoudy91 
Hi! Can I ask about what happened to this one and why it wasn't merged? As far as I can tell there is currently no way to access concepts with CMA, so this is a big need!

luizfonseca:
Hi @cgcompassion, we will be closing this PR as there's more work to be done for us to enable it via the AppSDK (concepts/schemes are bound to the org, but apps are bound to spaces). Just enabling the types unfortunately won't make it work and you wouldn't be able to interact with any of the methods (we need to allow those in multiple places).

I don't have a explicit timeline for when this might come as this require some changes in how the AppSDK handles some of the CMA calls and how the API is exposed, but we will announce in a changelog when it does.

PR #2130, opened 2025-01-27
fix: force release to fix type conflicts with old contentful-management

MayaGillilan:


colomolo:
Not sure if no changes will produce the new release, but let's try

colomolo (approved):

MayaGillilan:
> Not sure if no changes will produce the new release, but let's try

Also not sure 🤔 we can also just add a space in somewhere or something I suppose

PR #2284, opened 2025-07-15
feat(api): add release parameter to entry API and update createEntry [EXT-6548]

lewisjcs:
# Purpose of PR

- Enhanced makeEntryAPI to accept a release parameter.
- Updated createEntry function to handle the release context, throwing errors for unsupported publish/unpublish actions in release mode.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/lib/entry.ts
+++ b/lib/entry.ts
@@ -1,6 +1,6 @@
 import { Channel } from './channel'
 import { MemoizedSignal } from './signal'
-import { EntryAPI, EntryFieldInfo, EntrySys, Metadata, TaskAPI } from './types'
+import { EntryAPI, EntryFieldInfo, EntrySys, Metadata, Release, TaskAPI } from './types'
 import { ExhaustiveEntryFieldAPI } from './types/field.types'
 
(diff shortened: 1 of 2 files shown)


lib/entry.ts
```diff
@@ -16,6 +16,7 @@ export default function createEntry(
   entryData: any,
   fieldInfo: EntryFieldInfo[],
   createEntryField: (info: EntryFieldInfo) => ExhaustiveEntryFieldAPI,
+  release?: Release,
```
jsdalton, line 19: I think we should not depend on the release itself (acting as a "scope") to determine whether entryData is/is not a "release entry". this information is "extrinsic" to the function itself -- meaning all it takes is for someone to call this function incorrectly and it'll falsely assume the publish/unpubllsh are not valid.

jsdalton: After chatting with Tolkien w've learned that release entries are indistinguishable from regular entries in the data themselves, and we rely on passing / preserving release context to make this distinction wherever it's needed.

So I take back the earlier comment and we can proceed with the general approach.

jsdalton (approved):

PR #2287, opened 2025-07-17
Feat/release id [EXT-6547]

lewisjcs:
# Purpose of PR

This pull request enhances the `ui-extensions-sdk` to support the "Releases v2" project by ensuring apps are aware of the release context they are operating in. This change specifically addresses ticket [EXT-6547](https://contentful.atlassian.net/browse/EXT-6547).

We are adding the `releaseId` to the `sdk.ids` object when an application is running within a release. This provides developers with a straightforward way to access the current release's ID, enabling them to build release-aware features.

A primary use case for this is to allow an app to fetch data or modify its behavior based on the specific release it's interacting with. This is a foundational requirement for many of the planned features in Releases v2.

## PR Checklist

- [x] Tests are passing

[EXT-6547]: https://contentful.atlassian.net/browse/EXT-6547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -64,6 +64,8 @@ function makeSharedAPI(
   const { user, parameters, locales, ids, initialContentTypes, hostnames, release } = data
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
+  const idsWithRelease = release ? { ...ids, release: release.sys.id } : ids
+
   return {
     cma: createCMAClient(ids, channel),
     cmaAdapter: createAdapter(channel),
@@ -89,7 +91,7 @@ function makeSharedAPI(
       error: (message: string) => channel.send('notify', { type: 'error', message }),
       warning: (message: string) => channel.send('notify', { type: 'warning', message }),
     },
-    ids,
+    ids: idsWithRelease,
     hostnames,
     release,
     access: {
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -64,7 +64,7 @@ function makeSharedAPI(
   const { user, parameters, locales, ids, initialContentTypes, hostnames, release } = data
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
-  const idsWithRelease = release ? { ...ids, release: release.sys.id } : ids
+  const effectiveIds = release ? { ...ids, release: release.sys.id } : ids
 
   return {
     cma: createCMAClient(ids, channel),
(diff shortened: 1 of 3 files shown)


lib/api.ts
```diff
@@ -64,6 +64,8 @@ function makeSharedAPI(
   const { user, parameters, locales, ids, initialContentTypes, hostnames, release } = data
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
+  const idsWithRelease = release ? { ...ids, release: release.sys.id } : ids
```
BobHemphill76, line 67:
```suggestion
  const maybeIdsWithRelease = release ? { ...ids, release: release.sys.id } : ids
```

lib/api.ts
```diff
@@ -89,7 +91,7 @@ function makeSharedAPI(
       error: (message: string) => channel.send('notify', { type: 'error', message }),
       warning: (message: string) => channel.send('notify', { type: 'warning', message }),
     },
-    ids,
+    ids: idsWithRelease,
```
BobHemphill76, line 94:
```suggestion
    ids: maybeIdsWithRelease,
```

lib/cma.ts
```diff
@@ -13,6 +13,8 @@ export function createCMAClient(ids: IdsAPI, channel: Channel): CMAClient {
         environmentId: ids.environmentAlias ?? ids.environment,
         spaceId: ids.space,
         organizationId: ids.organization,
+        // TODO: Shipped in the CMA
```
BobHemphill76, line 16: I think this is a public repo, we might want to remove this TODO

BobHemphill76 (changes requested):
We should probably remove the TODO here.

tylerwashington888 (approved):

lib/api.ts
```diff
@@ -64,6 +64,8 @@ function makeSharedAPI(
   const { user, parameters, locales, ids, initialContentTypes, hostnames, release } = data
   const currentLocation = data.location || locations.LOCATION_ENTRY_FIELD
 
+  const effectiveIds = release ? { ...ids, release: release.sys.id } : ids
```
jsdalton, line 67: I'm not against this approach per se -- but I wonder why we append the `releaseId` here in the SDK instead of doing so upstream ie in the channel sender that provides the `data` attribute (where the other `ids` are).

BobHemphill76 (approved):

PR #2297, opened 2025-07-25
feat: api, navigator): add optional release parameter to createNavigator so openEntry method could work properly.

tylerwashington888:
when an app calls sdk.navigator.openEntry(), in order for us to keep the user in the context of the current release when the entry the app wants to navigate to is in the release we need to:

1. allow skd.navigator api module to become release aware
2. create a function called entryInRelease() that checks to see if the entry being opened is in the release or not so that we can pass a entryInRelease value back up the postMessageChannel for the host app to handle.

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -83,7 +83,7 @@ function makeSharedAPI(
     space: createSpace(channel, initialContentTypes),
     dialogs: createDialogs(channel, ids),
     // Typecast because promises returned by navigator methods aren't typed
-    navigator: createNavigator(channel, ids, release || undefined) as NavigatorAPI,
+    navigator: createNavigator(channel, ids, release) as NavigatorAPI,
     notifier: {
       success: (message: string) => channel.send('notify', { type: 'success', message }),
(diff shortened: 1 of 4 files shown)


lib/api.ts
```diff
@@ -83,7 +83,7 @@ function makeSharedAPI(
     space: createSpace(channel, initialContentTypes),
     dialogs: createDialogs(channel, ids),
     // Typecast because promises returned by navigator methods aren't typed
-    navigator: createNavigator(channel, ids) as NavigatorAPI,
+    navigator: createNavigator(channel, ids, release || undefined) as NavigatorAPI,
```
jjolton-contentful, line 86: surprised to see this `|| undefined` is required since release is an optional property on ConnectMessage

jjolton-contentful (approved):

lib/api.ts
tylerwashington888: fixed

jjolton-contentful (approved):

PR #2582, opened 2026-06-01
feat: add ExO Experience Editor Toolbar App SDK surface [EXT-7361]

jjolton-contentful:
[EXT-7361](https://contentful.atlassian.net/browse/EXT-7361) · Epic: [EXT-6925](https://contentful.atlassian.net/browse/EXT-6925)

## Summary

Introduces the **ExO (Experience Orchestration) App SDK surface** — the first App SDK support for apps that run in the Experience Editor Toolbar location. The change is back-compatible: it adds a new `sdk.exo` namespace, the `experience-toolbar` location constant, and the supporting public types. No existing surface is changed or removed, though several existing types gain members (`ConnectMessage` gains an optional `exo` field, the locations map and `KnownAppSDK` union each gain the new location/SDK).

- **`sdk.exo` namespace** (`lib/exo.ts`): `context` (`{ type: 'experience' | 'fragment'; entityId }`), `onContextChanged()`, `getUiMode()` / `onUiModeChanged()` (`'form' | 'visual'`), and `experience` (snapshot/nodes/selection/Data Assembly access) — wired over the existing iframe `Channel` protocol.
- **`LOCATION_EXPERIENCE_TOOLBAR` = `'experience-toolbar'`** (`lib/locations.ts`): a single location constant serving both Experience and Fragment editing contexts, discriminated at runtime via `sdk.exo.context.type`.
- **`ExperienceEditorToolbarAppSDK`** and the ExO type surface exported from the package entrypoint (`ExoSDK`, `ExoContext`, `ExoNodeAPI`, `ExoSelectionAPI`, `ExperienceAPI`, `DataAssemblyAPI`, binding/property/Data-Assembly types).

Merging to `main` publishes these types as a **semver minor** (`@contentful/app-sdk` 4.57.0 → 4.58.0), unblocking the bridge and the host-side toolbar wiring.

## Availability / rollout note

Publishing these types does **not** make the location generally available. The `experience-toolbar` location is feature-flag gated at the App Definition validation layer and is currently restricted to internal Contentful apps. The types ship now so downstream packages can build against a stable, published contract; external developer-facing docs and templates follow under [EXT-6925](https://contentful.atlassian.net/browse/EXT-6925).

## Test plan

- [x] CI green: lint, type-check, tests on Node 22 + 24, build
- [x] Bundle-size check passes (gzipped estimate ~7977 B)
- [x] New unit coverage runs: `test/unit/exo.spec.ts` + `test/unit/api.spec.ts` additions
- [ ] Confirm squash-merge title keeps the `feat:` prefix so semantic-release publishes the **minor** 4.58.0 (no `BREAKING CHANGE:` footer / no `!` — surface is additive)
- [ ] Post-merge: `npm view @contentful/app-sdk@4.58.0 version` and confirm `dist-tags.latest` updated; spot-check exo exports resolve from the published package

[EXT-7361]: https://contentful.atlassian.net/browse/EXT-7361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXT-6925]: https://contentful.atlassian.net/browse/EXT-6925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXT-6925]: https://contentful.atlassian.net/browse/EXT-6925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lewisjcs (changes requested):
### 🔴 Must-fix before merge

1. **CI is red — lint fails on this PR's lines.** Reproduced on Node 24: `npm run lint` exits 1 with 6 `prettier/prettier` errors (`lib/types/api.types.ts:357–362`, `test/unit/api.spec.ts:263`). The `test` job dies at the lint step before reaching the tests. `npm run lint:fix` clears it. Blocking because release runs on merge-to-`main`.

2. **`exo` is typed non-optional but can be `undefined` at runtime.** `ExperienceEditorToolbarAppSDK` declares `exo: ExoSDK`, but `createExo()` returns `ExoSDK | undefined` and `makeExoAPI` assigns it straight through — the `without exo in handshake, sdk.exo is undefined` test confirms the path is reachable. A consumer casting to the public type and calling `sdk.exo.getUiMode()` gets a `TypeError` with no compile-time warning. Two options: throw when `exoInit` is absent (mirrors `createAgent`, keeps the non-optional type sound), or type the field `exo?: ExoSDK`. Throwing matches the existing convention. Fixing this after `4.58.0` ships is a breaking change.

3. **Internal ticket reference ships in the published `.d.ts`.** The `@see` JSDoc tag on `createExo` survives into `dist/exo.d.ts` (verified against a local build), and `package.json#files` ships `dist/**/*.d.ts` — so it lands in the public npm tarball. Drop the tag or replace it with a public docs link. A CI grep on `lib/` JSDoc would prevent recurrence.

### 🟠 Should-fix while the surface is still soft

- **Stub methods are about to freeze into the contract.** `getRootNodes()` always returns `[]` (callers can't distinguish "no nodes" from "not implemented"); `getNode()` is typed `ExoNodeAPI | null` but never returns null (forces a dead null-branch on every caller). Implement before publish or trim from the public `ExperienceAPI` until real — and drop the `| null` if it can't be null.
- **`getNode()` leaks channel handlers.** Each call runs `channel.addHandler('exo.nodeChanged.${nodeId}', …)` and drops the unsubscribe, so repeated calls on a long-lived session accumulate listeners. The `field-locale` pattern (construct-once, cache by id, reuse) is the precedent. The 5 connect-time handlers in `createExo` are fine — those match `agent`/`editor`.
- **`api.spec.ts` ExO-shape test is a no-op.** The `expect(...).to.have.all.keys(...)` sits behind `if (api.exo != null)`, but the shared `test()` helper sends no `exo`, so `api.exo` is `undefined` and the assertion never runs. It also lists 3 keys where `ExoSDK` has 5, so it would fail if it did run. The `exo.spec.ts` suite already covers the shape — delete it.
- **Naming/docs consistency.** `DataAssemblySDK` is the outlier next to `ExperienceAPI`/`ExoNodeAPI`/`ExoSelectionAPI` (`DataAssemblyAPI` matches). The 17-method `ExoNodeAPI` is light on per-member JSDoc vs siblings like `FieldAppSDK`. Both are cheap now, breaking later.

### 📝 PR description nits

- The `[ ] CI green` checkbox is currently inaccurate (see #1).
- "purely additive… no existing surface is modified" undersells it slightly — `ConnectMessage`, the locations map, and `KnownAppSDK` each gain members. Still back-compatible.
- Repo is public: consider genericizing the internal flag name and downstream-package reference in the rollout note, and dropping the generated-with trailer.

jjolton-contentful:
Verified every item against the code; all nine checked out. Pushed in 4984754.

**Must-fix**
1. Lint — fixed in the prior push (720bf6e); `npm run lint` exits 0 now.
2. `exo` soundness — `createExo` now throws `'Context data is required'` when the handshake lacks `exo`, mirroring `createAgent`. Keeps the public `exo: ExoSDK` non-optional and sound. Updated the two tests that asserted the old `undefined` behavior to assert the throw.
3. Internal ref — dropped the `@see EXT-7182` JSDoc; `grep '@see EXT-' lib/` is clean, so it no longer ships in `dist/*.d.ts`.

**Should-fix**
- `getNode()` leak — node APIs are now cached by id in `createExperienceAPI` (construct-once, reuse), so repeated calls don't accumulate `exo.nodeChanged.*` handlers. Added a regression test asserting the same instance + no extra `addHandler` call.
- No-op `api.spec.ts` test — deleted (it was dead behind the `if (api.exo != null)` guard, and would've thrown under the new behavior anyway; `exo.spec.ts` + the `ExO runtime` block cover the shape).
- `DataAssemblySDK` → `DataAssemblyAPI` — renamed across `exo.ts`, `exo.types.ts`, `index.ts`.
- JSDoc — added per-member docs to `ExoNodeAPI` and `DataAssemblyAPI`.

**Stubs (`getRootNodes()` / `getNode(): … | null`)** — keeping as-is; implemented for real in EXT-7189 before GA. Noted the contract risk; acceptable since the location is internal-only behind the flag until then.

**PR description** — updated: fixed the CI checkbox, softened the "additive" claim to name the `ConnectMessage`/locations-map/`KnownAppSDK` additions, genericized the internal flag/downstream reference, dropped the generated-with trailer.

Verification: `check-types`, `lint`, `test` (530 passing), `build`, `size` (7977 B) all green locally.

jjolton-contentful:
Verified both P1s against the canonical shapes; both confirmed and fixed in 548c05a.

**[P1] ResourceLink shape** — confirmed against `contentful-management`'s `ResourceLink<T>` (`common-types.d.ts`): the shape is `{ sys: { type: 'ResourceLink'; linkType: T; urn: string } }` — `linkType` + `urn`, no `id`. Our `DataAssemblyParameterValue` ResourceLink variant had `{ sys: { type: 'ResourceLink'; id } }`. Added a canonical `ResourceLink<T>` type (mirroring CM, exported from the entrypoint) and switched the variant to it. `setParameter()` now compiles against the correct payload.

**[P1] `ExperienceSnapshot.sys.template`** — confirmed against neighboring ExO code (`buildExperienceTree.ts` uses `ResourceLink<'Contentful:…'>`; the stories `resourceLink()` helper produces `{ sys: { type: 'ResourceLink', linkType, urn } }`). Changed `template` from `Link<'Template'>` to `ResourceLink<'Contentful:Template'>` and fixed the runtime default in `exo.ts`. Also gave `mockExperienceSnapshot` a realistic `template` — worth noting `tsconfig.json` excludes `test/` from `check-types`, so the mock's missing field never surfaced; that's also why the wrong published shape wasn't caught earlier.

**[P2] Stub tree methods** — leaving as-is per the call on the prior review round: `getNode()`/`getRootNodes()` stay in the contract and get real implementations in EXT-7189 before GA. The location is internal-only behind the flag until then.

Verification: `check-types`, `lint`, `test` (530 passing), `build`, `size` (7993 B) all green.

lewisjcs (dismissed):
### Re-review — must-fixes resolved

Re-ran all gates locally on Node 24: **lint, type-check, tests (530 passing), and build all green.**

Verified resolved:
- **Lint** — clean.
- **`exo` type soundness** — `createExo` now throws on missing handshake data (mirrors `createAgent`) and returns a non-optional `ExoSDK`; tests cover the throw path.
- **Internal ref** — gone from source and the built `dist/*.d.ts`.
- **`getNode()` handler leak** — id-keyed cache + a regression test asserting handler count stays flat on repeat calls.
- **Naming/JSDoc** — `DataAssemblyAPI` rename and per-member docs landed.

The schema-alignment reshapes (DesignValue, Data Assembly params, `ExoNodeType`) are good for GA.

Two optional, non-blocking cleanups while the surface is freezing:
- `test/unit/exo.spec.ts:414,417,439,441` — selection tests still use lowercase `nodeType: 'component'`; the rename to `'Component'` missed these (they pass only because the stub arg is untyped, so the literal skips the `ExoNodeType` check). Worth aligning so the tests don't assert a value the public type rejects.
- `lib/types/exo.types.ts:60` — `LinkType` looks orphaned after the Data Assembly reshape (no remaining consumer) but is still exported through the entrypoint. Candidate to drop before it freezes into the public surface.

Approving.

jjolton-contentful:
Both optional cleanups done in ad71ff5 — verified before changing:

- **Selection test casing** (`exo.spec.ts` 414/417/439/441) — aligned `nodeType: 'component'` → `'Component'`. Confirmed they passed only because the stub handler arg is untyped; now they assert a value the public `ExoNodeType` accepts.
- **Orphaned `LinkType`** — dropped from `exo.types.ts` and the entrypoint export. Confirmed no remaining consumer after the Data Assembly ResourceLink reshape (the DA types now use `'Contentful:Entry'` literals directly); `FieldLinkType` is unrelated and untouched.

All five gates green locally (check-types, lint, 530 tests, build, size). 

On `exoContext`: settled by code rather than the RFC — the host emitter (`user_interface` #31678, `exoToolbarContext.ts`) produces exactly `{ type: 'experience' | 'fragment'; entityId }` and uses `null` (not a `'none'` member) for no-context, matching the SDK. The RFC body's split-id/`'none'` form is the outlier; flagged for correction on the RFC and tracked in EXT-7362.

lewisjcs (approved):

PR #2589, opened 2026-06-09
fix: align ExO snapshot, bindings, and property descriptor with canonical schemas [EXT-7477]

jjolton-contentful:
[EXT-7477](https://contentful.atlassian.net/browse/EXT-7477)

## Summary
The ExO App Location SDK surface was designed in an earlier ExO phase and has drifted from the canonical entity model. Since no app consumes it yet, we have a window to correct it before adoption. This PR lands the corrections we've aligned on; the Data Assembly surface remains held pending review (tracked on EXT-7477).

- **`ExperienceSnapshot` → `Experience | Fragment` discriminated union.** `template` is now optional on the `Experience` arm and absent on the `Fragment` arm (which carries `componentType`), matching `experiences-api-schemas`. The old type could not represent a Fragment, forcing the host to hardcode `sys.type: 'Experience'`.
- **Dropped the stale `ExperienceBinding` / `'experience'` `BindingSourceType`** — an abandoned-design placeholder present in no schema and never wired up.
- **Removed the redundant `getBindingMetadata` method** — `getBinding` already returns the same information as a typed `Binding` union.
- **Fixed `ComponentPropertyDescriptor.binding` to use the `Binding` union.** Replaced the invented `ComponentPropertyBinding` (and the now-orphaned `BindingSourceType`) with the same `EntryBinding | ManualBinding` union `getBinding` returns, non-optional (`Binding | null`). Consumers filter with `p.binding?.type === 'entry'` and read `entryId`/`fieldId` off the `EntryBinding` arm — the shape the marketplace Component Sidebar flow needs.
- **Kept `area: 'content' | 'design'`.** Content-vs-design is a first-class ExO distinction (Bridge API Decision Log `#6`: "we need both"; modeled natively in the editor as `designPropertiesByNodeId`), and the toolbar RFC's property editor renders the two as separate sections via `properties.filter(p => p.area === ...)`. There's no reliable structural way to recover the distinction, so the discriminator stays.

Net descriptor shape: `{ key, area, value, binding: Binding | null }`.

## Notes
- **Semver: minor.** Technically type-contract changes, but the location is feature-flagged with zero consumers. Flagging explicitly since `@contentful/app-sdk` is a public contract.
- **Companion work, post-release** (these consume the published SDK, not this branch): example apps (experience-toolbar + experience-auditor) move their binding checks from `binding.sourceType` to `binding.type`; host bridge #31892 [EXT-7364] returns the `Binding`-typed descriptor and replaces hardcoded `type: 'Experience'` with `rootEntity.sys.type`.

## Test plan
- [x] `npm run check-types`
- [x] `npm run lint`
- [x] `npm test` (530 passing)
- [ ] Reviewer confirms the minor semver call for a public-but-unconsumed surface

[EXT-7477]: https://contentful.atlassian.net/browse/EXT-7477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXT-7364]: https://contentful.atlassian.net/browse/EXT-7364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lewisjcs (approved):

lib/types/exo.types.ts
```diff
@@ -21,18 +19,7 @@ export type ManualBinding = {
...
-}
-
-export type Binding = EntryBinding | ManualBinding | ExperienceBinding
-
-export interface ComponentPropertyBinding {
-  sourceType: BindingSourceType
-  entryId?: string
-  fieldId?: string
-}
+export type Binding = EntryBinding | ManualBinding
```
Chaoste, line 22: What kind of manual binding does this refer to? We only support entries AFAIK

lib/types/exo.types.ts
```diff
@@ -190,14 +184,29 @@ export interface ExoSelectionAPI {
...
+ * Mirrors the canonical `experiences-api-schemas` shapes — `Experience.sys.template`
+ * is optional there, and `Fragment.sys` carries a `componentType` ResourceLink and no template.
+ */
+export type ExperienceSnapshot =
+  | {
+      sys: {
+        id: string
+        type: 'Experience'
+        version: number
+        template?: ResourceLink<'Contentful:Template'>
```
Chaoste, line 199: For the record, it's just a left-over from a previous DB migration. An experience must always use a template, so we'll probably fix the types in the API soon to define template as required prop.

PR #2596, opened 2026-06-15
feat: expose available Data Assemblies via DataAssemblyAPI.getMany() [EXT-7479]

jjolton-contentful:
[EXT-7479](https://contentful.atlassian.net/browse/EXT-7479)

## Summary
- Adds a `getMany(): Promise<DataAssemblySummary[]>` discovery method to `DataAssemblyAPI`, so ExO toolbar apps can enumerate the Data Assemblies **available** in the current experience/fragment (and their parameter definitions) — not just the single active DA exposed by `get()`.
- Adds the public `DataAssemblySummary` type (`{ id, name, description?, parameters }`) and re-exports it from the package barrel.
- Wires the runtime method to a new `exo.getAvailableDataAssemblies` host↔app channel call.
- Purely additive, non-breaking → `feat:` minor on top of `4.58.2`.

## Design decisions (rationale; flagged for review)
- **Scope:** experience-scoped — the host already resolves this set (`dataAssembliesByDefiningEntityId`), so enumeration exposes what's resolved rather than adding an environment-wide CMA registry query.
- **Placement:** method on the existing `DataAssemblyAPI`, **not** a new plural `dataAssemblies` namespace — consistent with CMA's singular-namespace + plural-getter pattern (`cma.entry.getMany`).
- **Name:** `getMany()` to match the CMA SDK convention.
- **Type name:** the ticket originally proposed `AvailableDataAssembly`; renamed to **`DataAssemblySummary`** to follow this repo's noun+suffix style (sits beside `DataAssemblySnapshot`; "summary" = lighter listing view). A docs sweep found no formal upstream precedent for either name, so this was a free choice. The channel protocol string stays `exo.getAvailableDataAssemblies` (host↔app verb, independent of the SDK type).
- **⚠️ Return shape — flagged for pushback:** returns a **bare `DataAssemblySummary[]`**, NOT a CMA-style `{ items, total, skip, limit }` collection wrapper. The wrapper exists for pagination, which can't happen on this small in-editor set, and a bare array is internally consistent with the other `DataAssemblyAPI` methods. If a reviewer wants forward-compat optionality, switching to `{ items }` is a cheap follow-up.

**Ownership note:** experience-assembly is busy and signed off on merge-and-iterate, so this is reviewable by team-extensibility. Downstream host bridge and widget-renderer forwarding will ship in follow-ups — this PR is the SDK-contract-first step.

## Test plan
- [x] `npm run check-types` — clean
- [x] `npm run lint` — clean
- [x] `npm test` — 531 passing (incl. 2 new `.getMany()` specs)
- [x] `npm run build` — bundle created
- [x] `npm run size` — 7984 bytes gzipped (negligible delta)
- [ ] Reviewer: confirm the bare-array return shape is acceptable vs. a CMA collection wrapper

[EXT-7479]: https://contentful.atlassian.net/browse/EXT-7479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/types/exo.types.ts
+++ b/lib/types/exo.types.ts
@@ -93,9 +93,26 @@ export interface DataAssemblySnapshot {
   parameters: Record<string, DataAssemblyParameterDefinition>
 }
 
+/**
+ * A Data Assembly available within the current experience/fragment, with its parameter
+ * definitions. Returned by {@link DataAssemblyAPI.getMany}. Definition-only — no GraphQL
+ * resolvers, `return` mapping, or nested DAs (consistent with the trimmed DA surface).
+ */
+export interface DataAssemblySummary {
+  id: string
+  name: string
+  description?: string
+  parameters: Record<string, DataAssemblyParameterDefinition>
+}
+
 export interface DataAssemblyAPI {
   /** Returns the current Data Assembly snapshot for the active experience/fragment. */
   get(): DataAssemblySnapshot
+  /**
+   * Resolves the Data Assemblies available in the current experience/fragment, each with its
+   * parameter definitions. Discovery counterpart to {@link get}, which returns only the active DA.
+   */
+  getMany(): Promise<DataAssemblySummary[]>
   /** Resolves all Data Assembly parameter definitions, keyed by parameter id. */
   getParameters(): Promise<Record<string, DataAssemblyParameterDefinition>>
   /** Resolves a single parameter definition, or `null` if no parameter has that id. */
--- a/lib/exo.ts
+++ b/lib/exo.ts
@@ -13,6 +13,7 @@ import {
   ExoSelectionAPI,
   DataAssemblyAPI,
   DataAssemblySnapshot,
+  DataAssemblySummary,
   DataAssemblyParameterDefinition,
   DataAssemblyParameterValue,
   ComponentPropertyDescriptor,
@@ -229,6 +230,9 @@ function createDataAssemblyAPI(channel: Channel): DataAssemblyAPI {
   return {
     get(): DataAssemblySnapshot {
       return dataAssemblySignal.getMemoizedArgs()[0]
+    },
+    getMany(): Promise<DataAssemblySummary[]> {
+      return channel.call('exo.getAvailableDataAssemblies')
     },
     getParameters(): Promise<Record<string, DataAssemblyParameterDefinition>> {
       return channel.call('exo.getDataAssemblyParameters')
--- a/lib/types/index.ts
+++ b/lib/types/index.ts
@@ -156,6 +156,7 @@ export type {
   DataAssemblyParameterDefinition,
   DataAssemblyParameterValue,
   DataAssemblySnapshot,
+  DataAssemblySummary,
   DataAssemblyAPI,
   ExoNodeType,
   ExoNodeSnapshot,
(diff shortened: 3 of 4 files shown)


lib/types/exo.types.ts
```diff
@@ -93,9 +93,26 @@ export interface DataAssemblySnapshot {
...
+ * resolvers, `return` mapping, or nested DAs (consistent with the trimmed DA surface).
+ */
+export interface DataAssemblySummary {
+  id: string
+  name: string
+  description?: string
+  parameters: Record<string, DataAssemblyParameterDefinition>
+}
+
 export interface DataAssemblyAPI {
```
Chaoste, line 108: This API uses the same term "parameter" both when referring to definitions or values. I suggest making this explicit. Right now, `getParameter` and `setParameter` sound like they are related to the same concept but they are not. Suggested interface:

```
  getParameterDefinitions(): Promise<Record<string, DataAssemblyParameterDefinition>>
  getParameterValues(): Promise<Record<string, DataAssemblyParameterValue>>
  setParameterValues(updates: Partial<Record<string, DataAssemblyParameterValue>>): Promise<void>
```

lib/types/exo.types.ts
```diff
@@ -93,9 +93,26 @@ export interface DataAssemblySnapshot {
...
   /** Returns the current Data Assembly snapshot for the active experience/fragment. */
   get(): DataAssemblySnapshot
+  /**
+   * Resolves the Data Assemblies available in the current experience/fragment, each with its
+   * parameter definitions. Discovery counterpart to {@link get}, which returns only the active DA.
+   */
+  getMany(): Promise<DataAssemblySummary[]>
```
Chaoste, line 115: Following the existing `get()` pattern, `exoSdk.experience.dataAssembly.getMany()` sounds like I get _all data assemblies_ used on the experience. How about we make it more explicit that one doesn't confuse selected ones vs. available/compatible ones?

```suggestion
  getAllAvailable(): Promise<DataAssemblySummary[]>
```

Chaoste (changes requested):

lib/types/exo.types.ts
jjolton-contentful: Agreed, overloading "parameter" across definitions and values is ambiguous. 

Splitting the accessors to make the distinction explicit: `getParameterDefinitions()` / `getParameterDefinition(id)` for the definition side, and `setParameterValue(id, v)` / `setParameterValues(updates)` for the value side. 

Folding this into a broader reshape of the surface (see closing note) so the naming lands once rather than shifting again in a follow-up.

lib/types/exo.types.ts
jjolton-contentful: Good point, `get()` / `getMany()` reads as "one vs. all of the same thing," when the distinction here is active vs. available/compatible. Going with `getAvailable()` to make that explicit. Same note as above: this will land via the reshape rather than this PR.

jjolton-contentful:
Closing in favor of a follow-up that reshapes this surface holistically, the discovery method here lands alongside the parameter definition/value naming split and the `getMany` → `getAvailable` rename, so the public contract changes in one release rather than across two. Both review comments are carried into that work (EXT-7479 → EXT-7496). 

Appreciate the careful read on the naming.

PR #2611, opened 2026-06-26
feat(dialogs): add cross-space resource entity selectors and recommendations support

david-shibley-contentful:
## Summary
- Adds `selectSingleResourceEntity` and `selectMultipleResourceEntities` to `DialogsAPI` for cross-space reference selection — these methods are called by the Rich Text Editor v6.3.9+ when inserting resource hyperlinks and embedded resource blocks
- Adds `recommendations` field to `selectSingleEntry` / `selectMultipleEntries` options to support the `/semantic/reference-suggestions` call used for AI-assisted entry suggestions
- Adds new exported types: `ResourceLink`, `ResourceEntitySelectorOptions`, `AllowedResource`, `RecommendationsOptions`

**Context:** The Rich Text Versioning app breaks with `@contentful/field-editor-rich-text` ≥ 6.3.9 because those editor versions call `sdk.dialogs.selectSingleResourceEntity(...)` and pass `recommendations: { searchQuery }` to `selectSingleEntry(...)`, neither of which existed in the App SDK contract. A previous app-local stub workaround ([contentful/apps#10987](https://github.com/contentful/apps/pull/10987)) was reverted after customer issues — this is the proper SDK-level fix.

## Test plan
- [x] TypeScript build passes (`npm run build`)
- [x] All 455 existing tests pass (`npm test`)
- [ ] Verify `selectSingleResourceEntity` is called correctly by the Rich Text Editor in an app context (cross-space resource hyperlink flow)
- [ ] Verify `recommendations.searchQuery` is forwarded correctly to the host via the channel when inserting an entry hyperlink

### DIFF
--- a/lib/dialogs.ts
+++ b/lib/dialogs.ts
@@ -5,6 +5,8 @@ import {
   OpenAlertOptions,
   OpenConfirmOptions,
   IdsAPI,
+  ResourceEntitySelectorOptions,
+  ResourceLink,
 } from './types'
 
 const isObject = (o: any) => typeof o === 'object' && o !== null && !Array.isArray(o)
@@ -28,6 +30,10 @@ export default function createDialogs(channel: Channel, ids: IdsAPI): DialogsAPI
     selectMultiplePatterns: openEntitySelector.bind(null, 'Pattern', true),
     selectSingleComponentDefinition: openEntitySelector.bind(null, 'ComponentDefinition', false),
     selectMultipleComponentDefinitions: openEntitySelector.bind(null, 'ComponentDefinition', true),
+    selectSingleResourceEntity: (options?: ResourceEntitySelectorOptions) =>
+      openResourceEntitySelector(false, options) as Promise<ResourceLink | null>,
+    selectMultipleResourceEntities: (options?: ResourceEntitySelectorOptions) =>
+      openResourceEntitySelector(true, options) as Promise<ResourceLink[] | null>,
   }
 
   function openSimpleDialog(type: string, options?: OpenAlertOptions | OpenConfirmOptions) {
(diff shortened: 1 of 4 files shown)


jjolton-contentful:
Code clean and contract correct. Well-scoped, additive contract-fill that closes a real gap 🚀 Just one note and one question from my end. 

`test/unit/dialogs.spec.ts` is a table-driven spec, every sibling selector (`selectSingleEntry`, `selectSingleComponentDefinition`, etc.) is registered in the `entitySelectorMethods` table. The two new methods use a different helper (`openResourceEntitySelector`, which sets `opts.multiple` and uses dialog type `resourceEntitySelector`), so they're untested. IMO worth adding a case asserting they call `openDialog` with `resourceEntitySelector` and the right multiple flag.

The PR types `type: string` and `source: string`, but the RTE's actual `AllowedResource` uses the literal type: `'Contentful:Entry'` (see https://github.com/contentful/field-editors/blob/master/packages/rich-text/src/helpers/getAllowedResourcesForNodeType.ts#L7). Bare string is forward-compatible (other resource types may come) and structurally accepts the RTE's payload, so it works, but it's worth a deliberate decision: do we want to signal "Entry only for now" or keep it open for future resource types?

david-shibley-contentful:
Tests added — both methods now have table-driven coverage asserting they call openDialog with resourceEntitySelector and the correct multiple flag. 536 passing.

On the type question: deliberate call to keep it string. Field-editors uses 'Contentful:Entry' today, but the SDK ships
independently and the cross-space reference spec doesn't preclude other resource types from being added in the future. Narrowing to a literal here would be a breaking change the moment a second type shows up, and we'd be baking field-editors' current implementation detail into the App SDK contract.

If the team has a strong signal that Entry is the only type ever, happy to tighten it — but string feels like the safer posture for a public API.

jjolton-contentful (approved):

PR #2624, opened 2026-07-06
fix: correct SlotDescriptor to allowedResources[] shape [EXT-7477]

jjolton-contentful:
[EXT-7477](https://contentful.atlassian.net/browse/EXT-7477)

## Summary
- `SlotDescriptor.allowedComponentTypeIds: string[]` did not match the canonical component-type slot schema. Slots carry structured `allowedResources` (`{ type: 'Contentful:ComponentType'; source; allowedTypes: string[] }`), not a flat id list. This was flagged during the EXT-7477 `sdk.experiences` surface review.
- Replaces the field with `allowedResources: SlotAllowedResource[]` and exports the new `SlotAllowedResource` type.
- The ExO toolbar location is **pre-GA with no external consumers**, so this contract change is safe to make now.

## Coordination note
The host bridge's `getNodeSlotDescriptor` already resolves `allowedResources[].allowedTypes` and was flattening it purely to satisfy the old flat type. Once this lands, the host can pass the structured shape through unchanged. The host-side return shape must be aligned to match this in the toolbar-bridge PR, so this SDK type and that host change should be verified together before GA.

## Test plan
- [x] `npm run check-types`
- [x] `npm run lint`
- [x] `npm test` (540 passing)
- [x] `npm run build`
- [x] `npm run size` (8310 bytes gzipped)
- [ ] Confirm host `getNodeSlotDescriptor` returns the matching `allowedResources[]` shape before GA

[EXT-7477]: https://contentful.atlassian.net/browse/EXT-7477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/types/experience.types.ts
+++ b/lib/types/experience.types.ts
@@ -134,9 +134,16 @@ export interface ExperienceNodeSnapshot {
   nodeType: ExperienceNodeType
 }
 
+/** One allowed resource for a slot: the Component(s) that may be placed in it. */
+export interface SlotAllowedResource {
+  type: 'Contentful:ComponentType'
+  source: string
+  allowedTypes: string[]
+}
+
 export interface SlotDescriptor {
   id: string
-  allowedComponentTypeIds: string[]
+  allowedResources: SlotAllowedResource[]
   currentItems: Array<{ nodeId: string; nodeType: 'Fragment' | 'InlineFragment' }>
 }
 
--- a/lib/types/index.ts
+++ b/lib/types/index.ts
@@ -162,6 +162,7 @@ export type {
   DataAssemblyAPI,
   ExperienceNodeType,
   ExperienceNodeSnapshot,
+  SlotAllowedResource,
   SlotDescriptor,
   ExperienceSnapshot,
   ExperienceNodeAPI,


lib/types/experience.types.ts
```diff
@@ -134,9 +134,16 @@ export interface ExperienceNodeSnapshot {
   nodeType: ExperienceNodeType
 }
 
+/** One allowed resource for a slot: the Component(s) that may be placed in it. */
+export interface SlotAllowedResource {
+  type: 'Contentful:ComponentType'
```
BobHemphill76, line 139: Is there an RFC or any other documentation about this format?  we have CRNs as well as sys.type scattered all throughout our API that carry type enumerations.  I'm not sure I've ever seen this specific format, so just wondering where it's defined in our documentation?

BobHemphill76 (approved):

lib/types/experience.types.ts
jjolton-contentful: This is a mirror of the canonical component-type wire schema so the SDK type matches exactly what the host resolves. 

`type: 'Contentful:ComponentType'` is the existing `Contentful:<Entity> ResourceLink` discriminator (same one behind `ResourceLinkSchema('Contentful:DataAssembly' | 'Contentful:Fragment' | …)`), not a sys.type. 

`source` is a CRN, the schema actually pins it to the literal `crn:contentful:::experience:spaces/$self/environments/$self` but I left it as string in the SDK to avoid pulling schema-level constraints into the public types

PR #2625, opened 2026-07-06
feat: add experience metadata get/set surface [EXT-7501]

jjolton-contentful:
[EXT-7501](https://contentful.atlassian.net/browse/EXT-7501)

## Summary
- The Experience Toolbar SDK exposed no way to read, subscribe to, or mutate an experience/fragment's metadata (tags, taxonomy concepts, name), though the data already lives on the entity and the host already read-modify-saves it internally.
- Adds `ExperienceMetadata` (tags + concepts as classic `Link`s, plus `name`), threads optional `metadata` onto both `ExperienceSnapshot` arms (Experience *and* Fragment), and adds `getMetadata()` / `setMetadata(patch)` / `onMetadataChanged(cb)` to `ExperienceAPI`.
- `getMetadata()` reads synchronously off the current snapshot and returns `ExperienceMetadata | undefined` — matching the `getMetadata()` contract on the entry/asset locations (`undefined` when the entity carries no metadata).
- `onMetadataChanged(cb)` mirrors entry/asset parity. It derives off the existing experience snapshot signal, firing immediately then only when the metadata actually changes; snapshots that bump only `sys` are ignored.
- `setMetadata(patch)` sends a partial update over the channel and returns `Promise<void>` so an app can await host acceptance before `save()` — matching the other mutating calls (`save`/`publish`/`setDesignProperty`) rather than fire-and-forget.
- Partial-merge semantics: only keys present in `patch` change; a provided `tags`/`concepts` array replaces that field wholesale (pass `[]` to clear), an omitted key is left untouched.
- Additive and pre-GA — no external consumers.

## Scope note
This is **Layer 1** (public SDK surface) of EXT-7501. Layers 2 (`notSupported` stub) and 3 (host bridge read/write) are deliberately deferred: the graceful-degradation stub only matters once the host method exists. This PR settles the public contract those layers implement against. The host `setMetadata` handler + `metadata` on the emitted snapshot land in the toolbar-bridge PR before GA.

## Test plan
- [x] `npm run check-types`
- [x] `npm run lint`
- [x] `npm test` (548 passing)
- [x] `npm run build`
- [x] `npm run size` (8388 bytes gzipped)
- [ ] Host `setMetadata` + snapshot `metadata` implemented in the landing toolbar-bridge PR before GA

[EXT-7501]: https://contentful.atlassian.net/browse/EXT-7501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/experience.ts
+++ b/lib/experience.ts
@@ -107,11 +107,23 @@ function createExperienceAPI(channel: Channel, initial?: ExperienceSnapshot): Ex
     onChange(cb: (v: ExperienceSnapshot) => void): Unsubscribe {
       return experienceSignal.attach(cb)
     },
-    getMetadata(): ExperienceMetadata {
-      return experienceSignal.getMemoizedArgs()[0].metadata ?? {}
+    getMetadata(): ExperienceMetadata | undefined {
+      return experienceSignal.getMemoizedArgs()[0].metadata
     },
     setMetadata(patch: Partial<ExperienceMetadata>): Promise<void> {
       return channel.call('exo.setExperienceMetadata', patch)
+    },
+    onMetadataChanged(cb: (metadata?: ExperienceMetadata) => void): Unsubscribe {
+      // Diff-guard the memoized snapshot so sys-only churn doesn't fire a spurious metadata change.
+      let previous: string | undefined
+      let seeded = false
+      return experienceSignal.attach((snapshot) => {
+        const next = JSON.stringify(snapshot.metadata)
+        if (seeded && next === previous) return
+        seeded = true
+        previous = next
+        cb(snapshot.metadata)
+      })
     },
     save(): Promise<void> {
       return channel.call('exo.saveExperience')
--- a/lib/experience.ts
+++ b/lib/experience.ts
@@ -6,6 +6,7 @@ import {
   UiMode,
   Unsubscribe,
   ExperienceSnapshot,
+  ExperienceMetadata,
   ExperienceAPI,
   ExperienceNodeAPI,
   ExperienceNodeSnapshot,
@@ -105,6 +106,12 @@ function createExperienceAPI(channel: Channel, initial?: ExperienceSnapshot): Ex
     },
     onChange(cb: (v: ExperienceSnapshot) => void): Unsubscribe {
       return experienceSignal.attach(cb)
+    },
+    getMetadata(): ExperienceMetadata {
+      return experienceSignal.getMemoizedArgs()[0].metadata ?? {}
+    },
+    setMetadata(patch: Partial<ExperienceMetadata>): Promise<void> {
+      return channel.call('exo.setExperienceMetadata', patch)
     },
     save(): Promise<void> {
       return channel.call('exo.saveExperience')
--- a/lib/types/experience.types.ts
+++ b/lib/types/experience.types.ts
@@ -2,6 +2,8 @@
  * Experience SDK types.
  */
 
+import { Link } from './utils'
+
 export type Unsubscribe = () => void
 
 export type UiMode = 'form' | 'visual'
@@ -190,10 +192,21 @@ export interface ExperienceSelectionAPI {
 }
 
 /**
+ * Editable metadata carried by an experience or fragment: taxonomy `tags` and `concepts`,
+ * plus the entity `name`. `tags` and `concepts` are classic `Link`s (`sys.type: 'Link'`),
+ * distinct from the `ResourceLink`s used elsewhere in this surface.
+ */
+export interface ExperienceMetadata {
+  tags?: Link<'Tag', 'Link'>[]
+  concepts?: Link<'TaxonomyConcept', 'Link'>[]
+  name?: string
+}
+
+/**
  * Snapshot of the root entity being edited. Discriminated on `sys.type`:
  * an `Experience` (optionally backed by a template) or a `Fragment` (no template).
- * Mirrors the canonical `experiences-api-schemas` shapes — `Experience.sys.template`
- * is optional there, and `Fragment.sys` carries a `componentType` ResourceLink and no template.
+ * `Experience.sys.template` is optional, and `Fragment.sys` carries a `componentType`
+ * ResourceLink and no template.
  */
 export type ExperienceSnapshot =
   | {
@@ -203,6 +216,7 @@ export type ExperienceSnapshot =
         version: number
         template?: ResourceLink<'Contentful:Template'>
       }
+      metadata?: ExperienceMetadata
     }
   | {
       sys: {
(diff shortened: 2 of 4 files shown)


lib/experience.ts
```diff
@@ -106,6 +107,12 @@ function createExperienceAPI(channel: Channel, initial?: ExperienceSnapshot): Ex
     onChange(cb: (v: ExperienceSnapshot) => void): Unsubscribe {
       return experienceSignal.attach(cb)
     },
+    getMetadata(): ExperienceMetadata {
+      return experienceSignal.getMemoizedArgs()[0].metadata ?? {}
```
denkristoffer: Entry/Asset return `Metadata | undefined` and not `{}` when metadata is missing, should we stick to that pattern here?

denkristoffer (commented):
SDK precedent for entries/assets is that metadata is delivered with the initial entity state and exposed through `getMetadata()`/`onMetadataChanged()`, not fetched separately. Will we keep that convention here?

jjolton-contentful:
Thanks for the review @denkristoffer — both good calls, addressed in 4d1632c:

1. **`getMetadata()` return shape** — now `ExperienceMetadata | undefined` (dropped the `?? {}` coercion), matching the entry/asset `getMetadata()` contract exactly. Also removes the internal mismatch where `get()` could hand you `undefined` metadata while `getMetadata()` swore it was always there.
2. **Delivery convention** — metadata already rides on the initial snapshot + `exo.experienceChanged` (no separate fetch), so that half matched. Added `onMetadataChanged(cb)` for full parity with entry/asset — it derives off the existing `experienceSignal`, fires immediately then only when tags/concepts/name actually change (ignores sys-only snapshot churn), mirroring the sibling `onChange`/`onDesignPropertyChanged` subscriptions.

Verify green across the board (548 tests, +38 bytes gzip). Still Layer 1 / pre-GA — the host `setMetadata` + `metadata` on the emitted snapshot are the next pieces before this lights up end-to-end. Happy to pair on wiring those through if useful.

lib/experience.ts
```diff
@@ -106,6 +107,24 @@ function createExperienceAPI(channel: Channel, initial?: ExperienceSnapshot): Ex
...
+    },
+    setMetadata(patch: Partial<ExperienceMetadata>): Promise<void> {
+      return channel.call('exo.setExperienceMetadata', patch)
+    },
+    onMetadataChanged(cb: (metadata?: ExperienceMetadata) => void): Unsubscribe {
+      // Diff-guard the memoized snapshot so sys-only churn doesn't fire a spurious metadata change.
+      let previous: string | undefined
+      let seeded = false
+      return experienceSignal.attach((snapshot) => {
+        const next = JSON.stringify(snapshot.metadata)
```
primeinteger, line 121: Small thing on the diff-guard here: `JSON.stringify(snapshot.metadata)` is sensitive to key order and array order, so if the host ever reserializes metadata with a different key order -- or reorders a `tags`/`concepts` array without a semantic change -- we'd fire `cb` on a non-change. It's low risk in practice (a single serializer emits stable order, and the worst case is just an extra fire, never a missed one).

Two ways to go, whichever you prefer: leave it and continue relying on stable key order, or compare the fields explicitly (`tags`/`concepts` by id, `name` by value) if you'd rather be robust to reordering

tylerwashington888 (approved):

primeinteger (approved):
Approve. Clean, minimal, additive SDK surface with good tests and honest scope notes.

<details>
  <summary>Review findings</summary>

# Review: contentful/ui-extensions-sdk#2625

feat: add experience metadata get/set surface [EXT-7501] -- @jjolton-contentful

## Findings

No major findings. The change is additive, pre-GA, tightly scoped, and well-tested. One nice-to-have below, plus residual risks worth stating explicitly.

🟢 [lib/experience.ts:116](lib/experience.ts#L116)
- `onMetadataChanged` diff-guards via `JSON.stringify(snapshot.metadata)`, which is key-order and array-order sensitive. If the host ever reserializes the metadata object with a different key order (or reorders a `tags`/`concepts` array without a semantic change), the guard reports a spurious change and fires `cb`.
- Low risk: a single host serializer emits stable key order in practice, and the only failure mode is an extra fire (degrading to the pre-guard behavior), never a missed one.

## Remaining risks

- **`setMetadata` targets a host method that does not exist yet.** The PR is explicit that Layer 3 (host bridge) is deferred. `channel.call('exo.setExperienceMetadata', patch)` stores a resolver in `_responseHandlers` keyed by message id ([lib/channel.ts:46](lib/channel.ts#L46)) and only settles when the host posts a matching response. Until the host implements the method, calling `setMetadata` returns a **forever-pending promise** and leaks the response handler. Same shape applies to `getMetadata()`/`onMetadataChanged()` returning `undefined`/never-firing because the host does not yet populate `snapshot.metadata`. This is disclosed and acceptable for a pre-GA surface with no consumers, but it means the public API ships in a state where an app that adopts it early will hang silently. The deferred `notSupported` stub (Layer 2) is what will make this fail loudly -- worth making sure that lands before anything documents these methods as usable.
- **The reviewer's convention question looks addressed.** @denkristoffer asked whether metadata would be delivered with initial entity state and exposed via `getMetadata()`/`onMetadataChanged()` rather than fetched separately (the entry/asset precedent in [lib/entry.ts:74](lib/entry.ts#L74) / [lib/asset.ts:46](lib/asset.ts#L46)). Commit 4d1632c ("align experience metadata surface with entry/asset") does exactly that -- the read side is now synchronous off the snapshot, no separate fetch. Confirm that resolves the thread before merge.

## Test gaps

- Coverage is solid for the added surface: `getMetadata` absent/present, `setMetadata` channel call + promise pass-through, and `onMetadataChanged` initial fire / change / unchanged / detach.
- One untested edge: `onMetadataChanged` firing twice for two genuinely different metadata payloads in sequence (the tests cover initial -> changed and initial -> unchanged, but not changed -> changed-again where the guard must update `previous`). Cheap to add and would lock in the guard's state-update behavior.
- `getCall(2)` to grab the `exo.experienceChanged` handler couples the tests to handler registration order (context=0, uiMode=1, experience=2). Matches the existing style in this spec, so no change requested -- just noting it's a shared brittleness if handler order ever shifts.
</details>

lib/experience.ts
jjolton-contentful: Astute observation there. I'll accept the over-fire risk since I expect `ExperienceMetadata` might grow and wouldn't want `onMetadataChanged` to silently stop firing should someone forget to add new fields to the comparator

PR #2628, opened 2026-07-11
feat: wire getRootNodes() channel.call in createExperienceAPI [AIS-208]

lewisjcs:
# Purpose of PR

Wires `ExperienceAPI.getRootNodes()` to a real host round-trip. It now calls `channel.call('exo.getRootNodes')` and resolves the returned root-node snapshots into `ExperienceNodeAPI` instances, replacing the hard-coded `[]` stub that shipped no functional behavior.

`getRootNodes()`'s return type changes from synchronous `ExperienceNodeAPI[]` to `Promise<ExperienceNodeAPI[]>`. A shared `getOrCreateNodeApi` helper, extracted from `getNode()`, backs both methods so a node id already vended by `getNode()` returns the same cached instance from `getRootNodes()` rather than a duplicate — and registers only one `exo.nodeChanged.<id>` handler per node. When `getRootNodes()` resolves a node whose type disagrees with a stale `getNode()`-seeded cache entry, the cache entry is replaced with the fresh, authoritative type.

Rejections from `channel.call` propagate uncaught, consistent with `save()`/`publish()`.

Two known trade-offs: the return-type change ships as `feat:` rather than a major-version bump (deliberate — see rationale above), and the stale-cache-replacement test doesn't assert on handler count.

## PR Checklist

- [x] Tests are passing

> Note: This PR was created with the [contentful-github-create-pull-request](https://github.com/contentful/agents-kit/tree/main/libs/contentful-github-create-pull-request) skill, powered by [Agents Kit](https://github.com/contentful/agents-kit). To follow or use this workflow, see the [Agents Kit CLI skill docs](https://github.com/contentful/agents-kit/blob/main/apps/cli/README.md#consuming-skills).

### DIFF
--- a/lib/types/experience.types.ts
+++ b/lib/types/experience.types.ts
@@ -243,7 +243,8 @@ export interface ExperienceAPI {
   save(): Promise<void>
   publish(): Promise<void>
   getNode(nodeId: string): ExperienceNodeAPI | null
-  getRootNodes(): ExperienceNodeAPI[]
+  /** Resolves the experience/fragment root nodes from the host over the channel. */
+  getRootNodes(): Promise<ExperienceNodeAPI[]>
   selection: ExperienceSelectionAPI
   dataAssembly: DataAssemblyAPI
 }
--- a/lib/types/experience.types.ts
+++ b/lib/types/experience.types.ts
@@ -243,7 +243,6 @@ export interface ExperienceAPI {
   save(): Promise<void>
   publish(): Promise<void>
   getNode(nodeId: string): ExperienceNodeAPI | null
-  /** Resolves the experience/fragment root nodes from the host over the channel. */
   getRootNodes(): Promise<ExperienceNodeAPI[]>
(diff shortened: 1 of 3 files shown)


lib/types/experience.types.ts
```diff
@@ -243,7 +243,8 @@ export interface ExperienceAPI {
   save(): Promise<void>
   publish(): Promise<void>
   getNode(nodeId: string): ExperienceNodeAPI | null
-  getRootNodes(): ExperienceNodeAPI[]
+  /** Resolves the experience/fragment root nodes from the host over the channel. */
```
jjolton-contentful: nit: slightly self-evident comment here

jjolton-contentful (dismissed):

lib/types/experience.types.ts
lewisjcs: Fixed in 4df9b28

jjolton-contentful (approved):

